PR #26, opened 2020-05-24
Added papplMain feature

JaiLuthra1:
`delete` and `default`(set) respond successful-ok but fail to make changes.
Also, to allow a server to run, the system needs the driver callback to exist so had to use testpappl/pwg driver to fix it.
@michaelrsweet Could you suggest changes, if any?

### DIFF
--- a/pappl/main-private.h
+++ b/pappl/main-private.h
@@ -1,0 +1,59 @@
+//
+// Private main header file for the Printer Application Framework
+//
+// Copyright Â©Â 2020 by Michael R Sweet.
+//
+// Licensed under Apache License v2.0.  See the file "LICENSE" for more
+// information.
+//
+
+#ifndef _PAPPL_MAIN_PRIVATE_H_
+#  define _PAPPL_MAIN_PRIVATE_H_
+
+//
+// Include necessary headers
+//
+
+#  include "main.h"
+#  include "base-private.h"
+
+
+//
+// C++ magic...
+//
+
+#  ifdef __cplusplus
+extern "C" {
+#  endif // __cplusplus
+
+
+extern int    _papplMainAdd(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainCancel(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainDefault(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainDelete(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainJobs(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainListPrinters(void) _PAPPL_PRIVATE;
+extern int    _papplMainModify(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainOptions(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainServer(int num_options, cups_option_t *options, pappl_driver_cb_t cb) _PAPPL_PRIVATE;
+extern int    _papplMainShutdown(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainStatus(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainSubmit(int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PRIVATE;
+
+extern void   _papplMainAddPrinterURI(ipp_t *request, const char *printer_name, char *resource,size_t rsize) _PAPPL_PRIVATE;	
+extern http_t *_papplMainConnect(int auto_start) _PAPPL_PRIVATE;
+extern http_t *_papplMainConnectURI(const char *printer_uri, char  *resource, size_t rsize) _PAPPL_PRIVATE;
+extern char   *_papplMainGetDefaultPrinter(http_t *http, char *buffer, size_t bufsize) _PAPPL_PRIVATE;
+extern char   *_papplMainGetServerPath(char *buffer, size_t bufsize) _PAPPL_PRIVATE;
+extern void   _papplMainAddOptions(ipp_t *request, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+
+
+//
+// C++ magic...
+//
+
+#  ifdef __cplusplus
+}
+#  endif // __cplusplus
+
+#  endif // !_PAPPL_MAIN_PRIVATE_H
--- a/pappl/main-private.h
+++ b/pappl/main-private.h
@@ -1,7 +1,7 @@
 //
 // Private main header file for the Printer Application Framework
 //
-// Copyright Â©Â 2020 by Michael R Sweet.
+// Copyright ©� 2020 by Michael R Sweet.
 //
 // Licensed under Apache License v2.0.  See the file "LICENSE" for more
 // information.
@@ -27,24 +27,24 @@ extern "C" {
 #  endif // __cplusplus
 
 
-extern int    _papplMainAdd(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainCancel(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainDefault(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainDelete(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainJobs(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainListPrinters(void) _PAPPL_PRIVATE;
-extern int    _papplMainModify(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainOptions(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainServer(int num_options, cups_option_t *options, pappl_driver_cb_t cb) _PAPPL_PRIVATE;
-extern int    _papplMainShutdown(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainStatus(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern int    _papplMainSubmit(int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PRIVATE;
+extern bool    _papplMainAdd(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainCancel(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainDefault(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainDelete(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainJobs(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainListPrinters(char *base_name) _PAPPL_PRIVATE;
+extern bool    _papplMainModify(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainOptions(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainServer(char *base_name, int num_options, cups_option_t *options, pappl_driver_cb_t driver_cb, const char *cb_state, const char *footer, pappl_soptions_t soptions, int num_versions, pappl_version_t *sversion, pappl_contact_t *scontact, const char *geolocation, const char *organization) _PAPPL_PRIVATE;
+extern bool    _papplMainShutdown(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainStatus(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainSubmit(char *base_name, int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PRIVATE;
 
 extern void   _papplMainAddPrinterURI(ipp_t *request, const char *printer_name, char *resource,size_t rsize) _PAPPL_PRIVATE;	
-extern http_t *_papplMainConnect(int auto_start) _PAPPL_PRIVATE;
-extern http_t *_papplMainConnectURI(const char *printer_uri, char  *resource, size_t rsize) _PAPPL_PRIVATE;
+extern http_t *_papplMainConnect(char *base_name, int auto_start) _PAPPL_PRIVATE;
+extern http_t *_papplMainConnectURI(char *base_name, const char *printer_uri, char  *resource, size_t rsize) _PAPPL_PRIVATE;
 extern char   *_papplMainGetDefaultPrinter(http_t *http, char *buffer, size_t bufsize) _PAPPL_PRIVATE;
-extern char   *_papplMainGetServerPath(char *buffer, size_t bufsize) _PAPPL_PRIVATE;
+extern char   *_papplMainGetServerPath(char *base_name, char *buffer, size_t bufsize) _PAPPL_PRIVATE;
 extern void   _papplMainAddOptions(ipp_t *request, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
 
 
--- a/pappl/main-private.h
+++ b/pappl/main-private.h
@@ -27,21 +27,21 @@ extern "C" {
 #  endif // __cplusplus
 
 
-extern bool    _papplMainAdd(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainCancel(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainDefault(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainDelete(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainJobs(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainListPrinters(char *base_name) _PAPPL_PRIVATE;
-extern bool    _papplMainModify(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainOptions(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainServer(char *base_name, int num_options, cups_option_t *options, pappl_driver_cb_t driver_cb, const char *cb_state, const char *footer, pappl_soptions_t soptions, int num_versions, pappl_version_t *sversion, pappl_contact_t *scontact, const char *geolocation, const char *organization) _PAPPL_PRIVATE;
-extern bool    _papplMainShutdown(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainStatus(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
-extern bool    _papplMainSubmit(char *base_name, int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PRIVATE;
+extern bool    _papplMainAddPrinter(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainCancelJob(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainGetSetDefaultPrinter(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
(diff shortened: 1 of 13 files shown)


pappl/pappl-main.h
```diff
@@ -0,0 +1,73 @@
...
+#  define _PAPPL_MAIN_H_
+
+
+//
+// Include necessary headers
+//
+
+// for strlcpy
+#include "pappl-private.h"
+#include <spawn.h>
```
michaelrsweet, line 20: If "pappl-main.h" (renamed to "main.h") is a public header, it needs to include "pappl.h", not "pappl-private.h". Private definitions go in a header named "xxx-private.h" which will include "xxx.h" and potentially "pappl-private.h" or "base-private.h" as appropriate.

Since "<spawn.h>" is needed for the internal implementation of the code and not for the public API, that include belongs either in the private header (common to all of the main functions) or in the source file that uses it.

pappl/pappl-main.h
```diff
@@ -0,0 +1,73 @@
...
+
+// for driver
+#include "testsuite/testpappl.h"
+
+// Path to self
+char *pappl_path;
```
michaelrsweet, line 26: Library globals are generally a no-no. *if* you need this put is in a private header file, add "extern", prefix it with an underscore, use TitleCase naming (per the "CONTRIBUTING.md" file), annotate the visibility as private the the shared library, and place the descriptive comment *after* the definition for the code documentation tool. For example:

```
extern char *_papplPath _PAPPL_PRIVATE;  // Path to self
```

pappl/pappl-main.h
```diff
@@ -0,0 +1,73 @@
...
+// C++ magic...
+//
+
+#  ifdef __cplusplus
+extern "C" {
+#  endif // __cplusplus
+
+
+extern int    papplMain(int argc, char* argv[]) _PAPPL_PUBLIC;
+extern int    papplMainAdd(int num_options, cups_option_t *options) _PAPPL_PUBLIC;
```
michaelrsweet, line 39: As I mentioned in a previous comment, I think the sub-commands should all be private API (starting with an underscore and using _PAPPL_PRIVATE for visibility). For example:

```
extern int _papplMainAdd(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
```

pappl/pappl-main.h
```diff
@@ -0,0 +1,73 @@
...
+extern int    papplMainStatus(int num_options, cups_option_t *options) _PAPPL_PUBLIC;
+extern int    papplMainSubmit(int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PUBLIC;
+
+extern void   papplMainAddPrinterURI(ipp_t *request, const char *printer_name, char *resource,size_t rsize) _PAPPL_PUBLIC;	
+extern http_t *papplMainConnect(int auto_start) _PAPPL_PUBLIC;
+extern http_t *papplMainConnectURI(const char *printer_uri, char  *resource, size_t rsize) _PAPPL_PUBLIC;
+extern char   *papplMainGetDefaultPrinter(http_t *http, char *buffer, size_t bufsize) _PAPPL_PUBLIC;
+extern char   *papplMainGetServerPath(char *buffer, size_t bufsize) _PAPPL_PUBLIC;
+extern void   papplMainAddOptions(ipp_t *request, int num_options, cups_option_t *options) _PAPPL_PUBLIC;
+extern char	  *copy_stdin(char *name, size_t namesize) _PAPPL_PUBLIC;
```
michaelrsweet, line 58: These functions should be static to the source file using them or properly named as private API, e.g., "_papplCopyStdin", although I don't think copy_stdin needs to be available outside a single source file.

pappl/pappl-main.h
```diff
@@ -0,0 +1,73 @@
...
+extern int    papplMainServer(int num_options, cups_option_t *options) _PAPPL_PUBLIC;
+extern int    papplMainShutdown(int num_options, cups_option_t *options) _PAPPL_PUBLIC;
+extern int    papplMainStatus(int num_options, cups_option_t *options) _PAPPL_PUBLIC;
+extern int    papplMainSubmit(int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PUBLIC;
+
+extern void   papplMainAddPrinterURI(ipp_t *request, const char *printer_name, char *resource,size_t rsize) _PAPPL_PUBLIC;	
+extern http_t *papplMainConnect(int auto_start) _PAPPL_PUBLIC;
+extern http_t *papplMainConnectURI(const char *printer_uri, char  *resource, size_t rsize) _PAPPL_PUBLIC;
+extern char   *papplMainGetDefaultPrinter(http_t *http, char *buffer, size_t bufsize) _PAPPL_PUBLIC;
+extern char   *papplMainGetServerPath(char *buffer, size_t bufsize) _PAPPL_PUBLIC;
```
michaelrsweet, line 56: This should be configurable, perhaps as an argument to the papplMain function.

pappl/system-loadsave.c
```diff
@@ -107,7 +107,7 @@ papplSystemLoadState(
       pappl_printer_t	*printer;	// Current printer
 
 
-      if ((num_options = cupsParseOptions(value, 0, &options)) != 4 || (printer_id = cupsGetOption("id", num_options, options)) == NULL || atoi(printer_id) <= 0 || (printer_name = cupsGetOption("name", num_options, options)) == NULL || (device_id = cupsGetOption("did", num_options, options)) == NULL || (device_uri = cupsGetOption("uri", num_options, options)) == NULL || (driver_name = cupsGetOption("driver", num_options, options)) == NULL)
+      if ((num_options = cupsParseOptions(value, 0, &options)) != 5 || (printer_id = cupsGetOption("id", num_options, options)) == NULL || atoi(printer_id) <= 0 || (printer_name = cupsGetOption("name", num_options, options)) == NULL || (device_id = cupsGetOption("did", num_options, options)) == NULL || (device_uri = cupsGetOption("uri", num_options, options)) == NULL || (driver_name = cupsGetOption("driver", num_options, options)) == NULL)
```
michaelrsweet, line 110: Again, thanks for catching this!

testsuite/testpappl.c
```diff
@@ -50,6 +50,8 @@ int
 main(int  argc,				// I - Number of command-line arguments
      char *argv[])			// I - Command-line arguments
 {
+  papplMain(argc, argv);
```
michaelrsweet, line 53: Make a separate test program that uses the PWG test driver with papplMain. testpappl will eventually include a lot more unit test functionality beyond what papplMain offers.

JaiLuthra1:
@michaelrsweet The changes you requested have disappeared after a force-push due to renaming of the files. I have made the changes to not use testpappl/pwg-driver.c any more and make the entry points private wherever possible. For the system options, I think we can set some default options and use command line arguments to alter these based on user requirement. Should I do it this way?

pappl/main-private.h
```diff
@@ -0,0 +1,59 @@
...
+extern int    _papplMainListPrinters(void) _PAPPL_PRIVATE;
+extern int    _papplMainModify(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainOptions(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainServer(int num_options, cups_option_t *options, pappl_driver_cb_t cb) _PAPPL_PRIVATE;
+extern int    _papplMainShutdown(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainStatus(int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern int    _papplMainSubmit(int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PRIVATE;
+
+extern void   _papplMainAddPrinterURI(ipp_t *request, const char *printer_name, char *resource,size_t rsize) _PAPPL_PRIVATE;	
+extern http_t *_papplMainConnect(int auto_start) _PAPPL_PRIVATE;
```
michaelrsweet, line 44: I've been making obviously boolean arguments/return values use the `bool` type in PAPPL (a bit of an upgrade to the original LPrint code...)

pappl/main-subcommand.c
```diff
@@ -0,0 +1,1228 @@
...
+  // Get required values...
+  device_id    = cupsGetOption("device-id", num_options, options);
+  device_uri   = cupsGetOption("device-uri", num_options, options);
+  driver       = cupsGetOption("driver", num_options, options);
+  printer_name = cupsGetOption("printer-name", num_options, options);
+
+  if (!device_uri || !driver || !printer_name || !device_id)
+  {
+    if (!printer_name)
+      fprintf(stderr, "papplMain: Missing -d printer\n");
```
michaelrsweet, line 56: (Global comment) We need to pass in `basename(argv[0])` for all of the error messages, so that the printer application's name is used as the prefix instead of 'papplMain'. I suggest updating the prototypes for all of the sub-commands to accept a program name that papplMain() gets from basename(argv[0])...

pappl/main-subcommand.c
```diff
@@ -0,0 +1,1228 @@
...
+    if (!device_uri)
+      fprintf(stderr, "papplMain: Missing -v device-uri\n");
+    if (!driver)
+      fprintf(stderr, "papplMain: Missing -m driver\n");
+    if (!device_id)
+      fprintf(stderr, "papplMain: Missing -i device-id\n");
```
michaelrsweet, line 62: I think we should get the device ID from papplDeviceList instead of needing it from the command-line. An empty/NULL device ID string is also OK for papplPrinterCreate.

pappl/main-subcommand.c
```diff
@@ -0,0 +1,1228 @@
...
+  // Create the system object and run it...
+  if ((system = papplSystemCreate(soptions, system_name ? system_name : "PAPPL", hostname, port, "_print,_universal", cupsGetOption("spool-directory", num_options, options), logfile ? logfile : "-", loglevel, cupsGetOption("auth-service", num_options, options), /* tls_only */false)) == NULL)
+    return (1);
+
+  char sockname[1024];		// Socket filename
+
+  papplSystemAddListeners(system, NULL);
+  papplSystemAddListeners(system, _papplMainGetServerPath(sockname, sizeof(sockname)));
+  (*cb)(system);
+  papplSystemSetFooterHTML(system,
```
michaelrsweet, line 715: This needs to be configurable, and the default shouldn't be a PAPPL copyright otherwise we'll get lots of support requests in the future for somebody else's printer application that uses PAPPL... (speaking from experience on past projects...)

pappl/main-subcommand.c
```diff
@@ -0,0 +1,1228 @@
...
+                         "Copyright &copy; 2020 by Michael R Sweet. "
+                         "Provided under the terms of the <a href=\"https://www.apache.org/licenses/LICENSE-2.0\">Apache License 2.0</a>.");
+  papplSystemSetSaveCallback(system, (pappl_save_cb_t)papplSystemSaveState, (void *)"testpappl.state");
```
michaelrsweet, line 718: This also needs to be configurable.

pappl/main-subcommand.c
```diff
@@ -0,0 +1,1228 @@
...
+
+  if (!(val = cupsGetOption("clean", num_options, options)))
```
michaelrsweet, line 720: I don't think we need an option for this - it's useful for the test framework but a deployed printer application will likely have another way to "reset" the configuration with suitable warnings to the user.

pappl/main-subcommand.c
```diff
@@ -0,0 +1,1228 @@
...
+
+  return (0);
+}
+
+
+//
+// 'papplMainAddSystemOption()' - Add an option to the system.
+//
+
+void
```
michaelrsweet, line 1027: Maybe this can be passed to papplMain instead?

pappl/main.c
```diff
@@ -0,0 +1,840 @@
...
+static int    help(int status);
+
+
+//
+// 'papplMain()' - Main entry for pappl.
+//
+
+int           // O - Exit status
+papplMain(
+    int   argc,   // I - Number of command line arguments
```
michaelrsweet, line 37: Should have a way to pass in server options, software name/version (which gets copied to the system version number info), a callback for unknown sub-commands (to allow for expansion), a callback to show program usage, the name of the state file, and the driver callback data pointer. Maybe more - think about what you'd need when writing a printer application, what things you are likely to customize or would be unique to each printer application.

pappl/main.c
```diff
@@ -0,0 +1,840 @@
...
+
+#ifdef __APPLE__
+  if (!tmpdir)
+    tmpdir = "/private/tmp";
+#else
+  if (!tmpdir)
+    tmpdir = "/tmp";
+#endif // __APPLE__
+
+  snprintf(buffer, bufsize, "%s/pappl%d.sock", tmpdir, (int)getuid());
```
michaelrsweet, line 723: This needs to use a unique base name for each printer application, probably based on the printer application name (back to that `basename(argv[0])` I've been talking about...)

pappl/main.c
```diff
@@ -0,0 +1,840 @@
...
+  else
+    return ((int)n);
+}
+
+
+//
+// 'help()' - Show usage.
+//
+
+int           // O - Exit status
```
michaelrsweet, line 793: This needs to be a callback, or perhaps this calls the callback to add application-specific sub-commands and options?

pappl/main-private.h
```diff
@@ -0,0 +1,59 @@
...
+extern "C" {
+#  endif // __cplusplus
+
+
+extern bool    _papplMainAdd(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainCancel(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainDefault(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainDelete(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainJobs(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainListPrinters(char *base_name) _PAPPL_PRIVATE;
```
michaelrsweet, line 35: I think like the _papplMainActionTarget convention best, so can you update these to be consistent throughout?

pappl/main-private.h
```diff
@@ -0,0 +1,59 @@
...
+extern bool    _papplMainModify(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainOptions(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainServer(char *base_name, int num_options, cups_option_t *options, pappl_driver_cb_t driver_cb, const char *cb_state, const char *footer, pappl_soptions_t soptions, int num_versions, pappl_version_t *sversion, pappl_contact_t *scontact, const char *geolocation, const char *organization) _PAPPL_PRIVATE;
+extern bool    _papplMainShutdown(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainStatus(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainSubmit(char *base_name, int num_options, cups_option_t *options, int num_files, char **files) _PAPPL_PRIVATE;
+
+extern void   _papplMainAddPrinterURI(ipp_t *request, const char *printer_name, char *resource,size_t rsize) _PAPPL_PRIVATE;	
+extern http_t *_papplMainConnect(char *base_name, int auto_start) _PAPPL_PRIVATE;
```
michaelrsweet, line 44: "auto_start" should be type `bool`.

pappl/main.h
```diff
@@ -0,0 +1,48 @@
...
+
+//
+// Include necessary headers
+//
+
+#  include "base.h"
+#  include "system.h"
+
+
+typedef void (*pappl_driver_cb_t)(pappl_system_t *system);
```
michaelrsweet, line 22: This should already be defined in system.h, so it shouldn't be redefined here (particularly with a different prototype...)

pappl/main.h
```diff
@@ -0,0 +1,48 @@
...
+typedef void (*pappl_usage_cb_t)();
```
michaelrsweet, line 23: What's the difference between a usage callback and an error callback in this context?

Also, these need "void *data" arguments (at least) to provide a way for the caller to provide additional context.

Finally, if they are specific to the papplMain interface, use "pappl_main_foo_cb_t".

pappl/main.h
```diff
@@ -0,0 +1,48 @@
...
+//
+// C++ magic...
+//
+
+#  ifdef __cplusplus
+extern "C" {
+#  endif // __cplusplus
+
+
+extern int    papplMain(int argc, char* argv[], pappl_driver_cb_t driver_cb, const char *cb_state, const char *footer, pappl_soptions_t soptions, int num_versions, pappl_version_t *sversion, pappl_contact_t *scontact, const char *geolocation, const char *organization, pappl_usage_cb_t usage_cb, pappl_error_cb_t error_cb) _PAPPL_PUBLIC;
```
michaelrsweet, line 36: There needs to be a better way to initialize the system object, when needed. Perhaps a callback for creating the system object and setting values prior to the papplSystemRun call? Something like:

    typedef pappl_system_t *(*pappl_main_system_cb_t)(int num_options, cups_option_t *options, void *data);

pappl/main.h
JaiLuthra1: I just kept the two for expansion as you mentioned earlier.
Should I remove one?

pappl/main.h
```diff
@@ -0,0 +1,47 @@
...
+//
+// C++ magic...
+//
+
+#  ifdef __cplusplus
+extern "C" {
+#  endif // __cplusplus
+
+
+extern int    papplMain(int argc, char* argv[], pappl_main_usage_cb_t usage_cb, pappl_main_usage_cb_t subcommand_cb, pappl_main_system_cb_t system_cb) _PAPPL_PUBLIC;
```
michaelrsweet, line 35: Not sure about the subcommand_cb argument - if it is meant to be a callback for other sub-commands then it needs to get the options, files, and sub-command in addition to a void * data parameter.

Also, you need to provide the data point in papplMain, otherwise how does papplMain know what value to send to the callbacks?

pappl/main-private.h
```diff
@@ -0,0 +1,59 @@
...
+#  ifdef __cplusplus
+extern "C" {
+#  endif // __cplusplus
+
+
+extern bool    _papplMainAddPrinter(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainCancelJob(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainDefaultPrinter(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainDeletePrinter(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainShowJobs(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
```
michaelrsweet, line 34: ShowJobs -> ListJobs or ListPrinters -> ShowPrinters for consistency.

Also, what about a ShowDevices command?

pappl/main-private.h
```diff
@@ -0,0 +1,59 @@
...
+//
+
+#  ifdef __cplusplus
+extern "C" {
+#  endif // __cplusplus
+
+
+extern bool    _papplMainAddPrinter(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainCancelJob(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
+extern bool    _papplMainDefaultPrinter(char *base_name, int num_options, cups_option_t *options) _PAPPL_PRIVATE;
```
michaelrsweet, line 32: Probably this should be named _papplMainGetSetDefaultPrinter since it gets or sets the default printer, right?

pappl/main-private.h
JaiLuthra1: The show devices has been implemented in main.c by calling papplDeviceList without creating an additional function for the same.
@michaelrsweet Do we still need a separate function for it?

JaiLuthra1:
@michaelrsweet I have made the changes you suggested.

"showDevices" has already been implemented in main.c by calling papplDeviceList without creating an additional function for the same.
Do we still need a separate function for it?

pappl/main.c
```diff
@@ -0,0 +1,881 @@
...
+    if (!strcmp(argv[i], "--help"))
+    {
+      (*usage_cb)(data);
+    }
+    else if (!strcmp(argv[i], "--version"))
+    {
+      puts(PAPPL_VERSION);
+      return (0);
+    }
+    else if (!strcmp(argv[i], "--list-devices"))
```
michaelrsweet, line 82: Sub-commands should not use -- option prefix. Think git and other one-executable utilities.

michaelrsweet:
@JaiLuthra1 Sorry, I missed that (was reviewing from my iPhone...) I don't think it is a big deal either way, although from a purely design/maintenance perspective it is nicer to keep the same pattern and isolate the code accordingly.

Also, see my subsequent comment about not using "--sub-command" for sub-commands. Just as for lprint and Git, sub-commands should not be prefixed - only the options get a prefix.

pappl/main.c
```diff
@@ -325,49 +269,79 @@ papplMain(
...
+          printf("%s: Cannot print more files.\n", base_name);
+          return (1);
+        }
+        files[num_files++] = argv[i];
       }
-      files[num_files++] = argv[i];
     }
   }
 
-  if (!subcommand && num_files > 0)
```
JaiLuthra1: Had to get rid of this to differentiate between an unknown sub-command and file name.

michaelrsweet:
Cleanup is now complete. I did a little renaming (main -> mainloop) for clarity, added the missing "drivers" sub-command, and cleaned out all of the non-lprint options so things are focused on running as a standalone printer application and not as a unit test program (which we have separately...)

PR #28, opened 2020-06-01
Added Printer add/delete web interface

dsam82:
Michael, how to authenticate the hostname/ip-address string entered by the user as the pattern is different for both hostname and ip-address? Please suggest changes, if any.


michaelrsweet:
@dsam82 Sorry for the delay, what you can do is two-fold:

1. Use an input pattern that allows either an IPv4 address, an IPv6 address, or a hostname.
2. Once you start validating the form data in the "add" function, try to lookup the value with httpAddrGetList - if that fails, you can say so (set status string).

dsam82:
@michaelrsweet I have made the changes. Can you please review?

PR #30, opened 2020-06-14
Added log rotation and change log level functionality.

dsam82:
@michaelrsweet, the textarea on reload changes the cursor back to the start or the end of the text. How can we handle this?


michaelrsweet:
Don't use text area for the log, and consider appending new log data via Javascript and a resource callback that implements a REST endpoint for retrieving the newest log data.

pappl/system-webif.c
```diff
@@ -1405,6 +1537,119 @@ system_header(pappl_client_t *client,	// I - Client
...
+
+    // Open the log file
+    if ((log = cupsFileOpen(filename, "r")) == NULL)
+    {
+      papplLog(system, PAPPL_LOGLEVEL_DEBUG, "Unable to access log file (%s): %s", filename, strerror(errno));
+      return;
+    }
+
+    // Allocate memory for the logs
+    if ((logbuf = calloc(1, (size_t)loginfo.st_size + 1)) == NULL)
```
michaelrsweet, line 1613: Ouch, that could be a LOT of memory. Better to open the file, seek to the start position, and copy N bytes using a smallish (8k-64k) local buffer. Open the log file first with open (instead of cupsFileOpen), then fstat the file descriptor, then lseek, then a loop of read + httpWrite2 to send the log data, then close. That way you'll survive a log file rotation...

pappl/system.c
```diff
@@ -379,6 +385,31 @@ papplSystemRun(pappl_system_t *system)// I - System
   // Loop until we are shutdown or have a hard error...
   while (!shutdown_system)
   {
+    if (restart_logging && strcmp(system->logfile, "-"))
```
michaelrsweet, line 388: Check that system->logfile is not NULL, also for "syslog" (no rotation).

For simplicity this logic can be placed in the log rotate function (no-op) so that a SIGHUP is also the way to force a status summary to be logged...

pappl/log.c
```diff
@@ -250,6 +250,101 @@ papplLogPrinter(
...
+
+    // Rename old log file to new file name
+    if (rename(filename, backname) != 0)
+    {
+      fprintf(stderr, "error renaming already existing file %s: %s", filename, strerror(errno));
+      return;
+    }
+
+    // Create new log file
+    if ((system->logfd = open(filename, O_CREAT | O_WRONLY | O_APPEND | O_NOFOLLOW | O_CLOEXEC, 0600)) < 0)
```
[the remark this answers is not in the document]
dsam82: @michaelrsweet Should we add a function `papplSetLogFd()` in system-accessors.c when updating the logfd as we need to update the descriptor quite often in `papplLogCheck()` function?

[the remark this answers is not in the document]
michaelrsweet: @dsam82 No, since this is an internal bit of state and the log file is not mutable, just make sure that the appropriate logic is present in the function(s) that manipulate the logfd value (i.e. in log.c).

pappl/log.c
```diff
@@ -250,6 +250,101 @@ papplLogPrinter(
...
+  }
+}
+
+
+//
+// 'papplLogStart()' - Restart the logging
+//
+
+bool
+papplLogStart(
```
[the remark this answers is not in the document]
dsam82: I've added the log opening part if the log file is not present in the log rotate function but I am not sure how to add the log restart part in the function? Can you please elaborate?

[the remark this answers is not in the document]
michaelrsweet: @dsam82 If I was writing it, I'd use the following logic:

```
if log file open
    check size and rename to filename.O if over the limit
    close the log file
endif

open the log file
write status information to log file
```

pappl/system-webif.c
```diff
@@ -1405,6 +1537,121 @@ system_header(pappl_client_t *client,	// I - Client
...
+      return;
+    }
+
+    if (httpWriteResponse(client->http, code) < 0)
+      return;
+
+    // Read buffer and write to client
+    while(length > 0 && (bytes = read(fd, buffer, sizeof(buffer))) > 0)
+    {
+      if (length - bytes < 0)
```
michaelrsweet, line 1640: LGTM flags this - change to compare directly, e.g., `if (bytes > length)`.

pappl/log-private.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifndef _PAPPL_LOG_PRIVATE_H_
+#  define _PAPPL_LOG_PRIVATE_H_
+
+//
+//  inlcude necessary headers
+//
+
+#  include  "log.h"
+#  include "base-private.h"
```
michaelrsweet, line 18: Include base-private.h before log.h.

pappl/log-private.h
```diff
@@ -0,0 +1,27 @@
...
+
+//
+//  Functions
+//
+
+extern bool   _papplLogCheck(pappl_system_t *system) _PAPPL_PRIVATE;
+extern void   _papplPrinterIteratorInfoCallback(pappl_printer_t *printer, pappl_system_t *system) _PAPPL_PRIVATE;
```
michaelrsweet, line 25: Maybe rename to '_papplLogPrinterIteratorStatusCallback' to keep the namespace straight...

pappl/log.c
```diff
@@ -40,6 +41,156 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
 };
 
 
+//
+// '_papplLogCheck()' - Open/Rotate Log File
+//
+
+bool
```
michaelrsweet, line 48: Document what the return value means please...

pappl/log.c
```diff
@@ -40,6 +41,156 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+  }
+  else if (!strcmp(system->logfile, "-"))
+  {
+    // Log to stderr...
+    system->logfd = 2;
+    pthread_rwlock_unlock(&system->rwlock);
+
+    return (true);
+  }
+  else if (stat(system->logfile, &statbuf) < 0)
```
michaelrsweet, line 80: There is a problem with this logic - it assumes that system->logfd will be uninitialized if the log file doesn't exist, but that isn't correct.

pappl/log.c
```diff
@@ -40,6 +41,156 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+
+      return (false);
+    }
+
+    pthread_rwlock_unlock(&system->rwlock);
+
+  }
+
+  // Log the system status
+  gmtime_r(&system->start_time, &start_date);
```
michaelrsweet, line 136: And all of this status information isn't being logged to stderr or syslog. If you send SIGHUP to the application we want this sent out regardless of the log destination...

pappl/log.c
```diff
@@ -40,6 +41,156 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+  {
+    snprintf(resource, sizeof(resource), "/ipp/print/%s", printer->name);
+    for (resptr = resource + 11; *resptr; resptr ++)
+      if ((*resptr & 255) <= ' ' || *resptr == 0x7f)
+	      *resptr = '_';
+  }
+  else
+    strlcpy(resource, "/ipp/print", sizeof(resource));
+
+  papplLog(system, PAPPL_LOGLEVEL_INFO, "Printer '%s' at resource path '%s'.", printer->name, resource);
```
michaelrsweet, line 189: No need to "figure out" the resource path of the printer, it is provided in the "resource" member of the pappl_printer_t structure.

pappl/log.h
```diff
@@ -60,7 +60,8 @@ extern void		papplLogDevice(const char *message, void *data) _PAPPL_PUBLIC;
 extern void		papplLogJob(pappl_job_t *job, pappl_loglevel_t level, const char *message, ...) _PAPPL_PUBLIC _PAPPL_FORMAT(3, 4);
 extern void		papplLogPrinter(pappl_printer_t *printer, pappl_loglevel_t level, const char *message, ...) _PAPPL_PUBLIC _PAPPL_FORMAT(3, 4);
 extern void		papplLogSetCallback(pappl_system_t *system, pappl_logfunc_t cb, void *data) _PAPPL_PUBLIC;
-
+// extern bool   _papplLogCheck(pappl_system_t *system) _PAPPL_PUBLIC;
+// extern void   _papplPrinterIteratorInfoCallback(pappl_printer_t *printer, pappl_system_t *system) _PAPPL_PUBLIC;
```
michaelrsweet, line 64: Just remove these completely.

pappl/printer-private.h
```diff
@@ -123,6 +123,7 @@ extern void		_papplPrinterInitPrintDriverData(pappl_pdriver_data_t *d) _PAPPL_PR
 extern bool		_papplPrinterRegisterDNSSDNoLock(pappl_printer_t *printer) _PAPPL_PRIVATE;
 extern void		_papplPrinterUnregisterDNSSDNoLock(pappl_printer_t *printer) _PAPPL_PRIVATE;
 
+extern void   _papplPrinterIteratorInfoCallback(pappl_printer_t *printer, pappl_system_t *system) _PAPPL_PRIVATE;
```
michaelrsweet, line 126: As I noted in log-private.h, use a _papplLogXxx name for this one.

pappl/system-accessors.c
```diff
@@ -557,6 +557,29 @@ papplSystemGetLocation(
...
 
+//
+// 'papplSystemGetLogLevel()' - Get the system log level.
+//
+
+pappl_loglevel_t
+papplSystemGetLogLevel(
+    pappl_system_t *system)     // I - System
+{
+  return (system && (system->loglevel < PAPPL_LOGLEVEL_DEBUG || system->loglevel > PAPPL_LOGLEVEL_FATAL)) ? PAPPL_LOGLEVEL_UNSPEC : system->loglevel;
```
michaelrsweet, line 568: Enough to use `return (system ? system->loglevel : PAPPL_LOGLEVEL_UNSPEC);`

pappl/system-accessors.c
```diff
@@ -557,6 +557,29 @@ papplSystemGetLocation(
...
+
+//
+// 'papplSystemGetMaxLogSize()' - Get the system max log size
+//
+
+size_t
+papplSystemGetMaxLogSize(
+    pappl_system_t *system)     // I - System
+{
+  return (system && system->maxLogSize) ? system->maxLogSize : 0;
```
michaelrsweet, line 579: Enough to use `return (system ? system->maxLogSize : 0);`.

pappl/system.c
```diff
@@ -379,6 +385,16 @@ papplSystemRun(pappl_system_t *system)// I - System
...
```
michaelrsweet, line 388: Since the rotation logic is elsewhere and we want a status summary regardless of the log file, just check `if (restart_logging)` here

pappl/system.c
```diff
@@ -379,6 +385,16 @@ papplSystemRun(pappl_system_t *system)// I - System
...
+    {
+      restart_logging = false;
+      if (restart_logging && !system->logfile && strcmp(system->logfile, "syslog") && strcmp(system->logfile, "-"))
```
michaelrsweet, line 391: This code will never be executed since you just set restart_logging to false...

pappl/log.c
```diff
@@ -40,6 +41,156 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+    if (rename(system->logfile, backname) != 0)
+    {
+      fprintf(stderr, "error renaming already existing file %s: %s", system->logfile, strerror(errno));
+      return (false);
+    }
+
+    pthread_rwlock_wrlock(&system->rwlock);
+
+    // Close old log file
+    close(system->logfd);
```
michaelrsweet, line 117: Unless you also do locking in the logging functions, you need to close the old file descriptor *after* you open the new one, which means making a copy of `system->logfd`:

```
int old_logfd = system->logfd;
system->logfd = open(...);
close(old_logfd);
```

pappl/log.c
```diff
@@ -40,6 +41,156 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+//
+// '_papplLogCheck()' - Open/Rotate Log File
+//
+
+bool
+_papplLogCheck(
+    pappl_system_t *system) // I - System
+{
+  struct stat statbuf;          // Log file information
+  size_t      maxLogSize;       // Maximum Log File Size
```
michaelrsweet, line 53: You can probably just reference `system->maxLogSize`.

pappl/system.c
dsam82: Oops, Sorry. My bad here.

pappl/log.c
```diff
@@ -40,6 +41,128 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+
+  // Log the system status information
+  gmtime_r(&system->start_time, &start_date);
+
+  papplSystemGetHostname(system, hostname, sizeof(hostname));
+
+  papplLog(system, PAPPL_LOGLEVEL_INFO, "Restarting log.");
+  papplLog(system, PAPPL_LOGLEVEL_INFO, "System started at %04d-%02d-%02dT%02d:%02d:%02dZ", start_date.tm_year + 1900, start_date.tm_mon + 1, start_date.tm_mday, start_date.tm_hour, start_date.tm_min, start_date.tm_sec);
+
+  if (hostname[0] == '/')
```
michaelrsweet, line 130: This should never be the case. The hostname is the network name, only a listen address will use a file path...

pappl/log.c
```diff
@@ -40,6 +41,128 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+    logfd = system->logfd;
+
+    // Create new log file
+    if ((system->logfd = open(system->logfile, O_CREAT | O_WRONLY | O_APPEND | O_NOFOLLOW | O_CLOEXEC, 0600)) < 0)
+    {
+      // Fallback to stderr if we can't open the log file...
+      perror(system->logfile);
+
+      system->logfd = 2;
+
```
michaelrsweet, line 114: Should still log the system status info; drop the "close" and "return" here and let it fall through...

pappl/log.c
```diff
@@ -40,6 +41,128 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+    system->logfd = -1;
+  }
+  else if (!strcmp(system->logfile, "-"))
+  {
+    // Log to stderr...
+    system->logfd = 2;
+  }
+
+  // Open the file in read mode
+  if ((logfd = open(system->logfile, O_RDONLY)) < 0)
```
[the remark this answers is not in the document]
dsam82: @michaelrsweet should I put the code related to file logging in an "else" after checking for syslog and stderr?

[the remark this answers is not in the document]
michaelrsweet: Yes.

pappl/log.c
```diff
@@ -40,6 +41,122 @@ static const int	syslevels[] =	// Mapping of log levels to syslog
...
+      // Check if file exists, if not create one
+      if ((system->logfd = open(system->logfile, O_CREAT | O_WRONLY | O_APPEND | O_NOFOLLOW | O_CLOEXEC, 0600)) < 0)
+      {
+        // Fallback to stderr if we can't open the log file...
+        perror(system->logfile);
+
+        system->logfd = 2;
+      }
+    }
+  }
```
michaelrsweet, line 87: All of the rotation code that follows also needs to be inside this "else" grouping...

dsam82: Sorry, I thought fstat will handle that.
I've modified it now.

dsam82:
@michaelrsweet Are there any other changes that need to be done?

PR #31, opened 2020-06-17
Add SNMP support

JaiLuthra1:


JaiLuthra1:
@michaelrsweet Do we need a find+callback mechanism here to connect to an snmp URI since the `addrname` is the DHCP IP address?

michaelrsweet:
@JaiLuthra1 I was thinking that the hostname in the "snmp://" URI could be the sysName OID value, or (if that isn't unique enough) maybe the MAC address from the Network MIB?

It definitely cannot just be an IP address because that is guaranteed to change. And we are really just supporting SNMP for network devices that don't support DNS-SD/mDNS (mainly label printers) so it would be even better if we can exclude SNMP-discovered printers that we've already found via DNS-SD.

JaiLuthra1:
@michaelrsweet Actually, sysname does give us the hostname that we use for the URI but we also have the current IP address in addrname that can be used for connecting. The URI still doesn't use the IP address.

michaelrsweet:
OK, that's fine. Lookup the matching SYSNAME and then use the response address to connect when opening.

JaiLuthra1:
@michaelrsweet I am not sure whether the `Port OID` used will work for all devices but that works with my HP printer. Also, I could not find any other Port OID to use.

pappl/device.c
```diff
@@ -1007,6 +1038,370 @@ pappl_dnssd_unescape(
...
+static int        // O - File descriptor or -1 on error
+pappl_snmp_connect(http_addr_t *addr)  // I - Device address
+{
+  int             fd = -1;      // File descriptor
+  http_addrlist_t *addrlist;    // Dummy address list
+
+
+  addr->ipv4.sin_port = htons(9100);
+
+  addrlist = calloc(1, sizeof(http_addrlist_t));
```
michaelrsweet, line 1067: If you are going to allocate this, you need to free it. But you can just have this as a local variable instead with no allocation...

pappl/device.c
```diff
@@ -1007,6 +1038,370 @@ pappl_dnssd_unescape(
...
+  memcpy(&(addrlist->addr), addr, sizeof(http_addr_t));
```
michaelrsweet, line 1068: You can also do "addrlist.addr = addr" since C supports copying of structs in assignments (gets internally mapped to a memcpy equivalent...)

pappl/device.c
```diff
@@ -1007,6 +1038,370 @@ pappl_dnssd_unescape(
...
+  _pappl_snmp_dev_t	*device,		// Matching device
+  	          *temp;			// New device entry
+  int                 i;    // Looping variable
+  const int           DevicePrinterOID[]   = { 1,3,6,1,2,1,25,3,1,5,-1 }; // OID for Printer
+  const int           SysNameOID[]         = { 1,3,6,1,2,1,1,5,0,-1 };
+  const int           HPDeviceIDOID[]      = { 1,3,6,1,4,1,11,2,3,9,1,1,7,0,-1 };
+  const int           LexmarkDeviceIdOID[] = { 1,3,6,1,4,1,641,2,1,2,1,3,1,-1 };
+  const int           ZebraDeviceIDOID[]   = { 1,3,6,1,4,1,10642,1,3,0,-1 };
+  const int           PWGPPMDeviceIdOID[]  = { 1,3,6,1,4,1,2699,1,2,1,2,1,1,3,1,-1 };
+  const int           PortOID[]      = { 1,3,6,1,2,1,43,14,1,1,9,1,5,-1 };
```
michaelrsweet, line 1281: I don't see this OID listed by any of my network printers...

.1.3.6.1.4.1.2699.1.2.1.3.1.1.6.1.1 is the PWG Printer Port Monitor MIB version.

.1.3.6.1.4.1.641.1.5.7.11.0 is a Lexmark-specific version.

.1.3.6.1.4.1.10642.20.10.20.15.2.1.10.1 is a Zebra-specific version.

.1.3.6.1.4.1.683.6.3.1.4.17.0 (SNMPv2-SMI::enterprises.683.6.3.1.4.17.0) appears to be a common OID for the TCP/IP raw port number with network label printers. See <http://oid-info.com/get/1.3.6.1.4.1.683.6.3.1.4.17>.

pappl/snmp-private.h
```diff
@@ -92,6 +92,14 @@ typedef struct _pappl_snmp_s		// SNMP data packet
   union _pappl_snmp_value_u object_value;	// object-value value
 } _pappl_snmp_t;
 
+enum				// Request IDs for each field
```
michaelrsweet, line 95: These are specific to the device code, so the enum should be defined there...

pappl/device.c
```diff
@@ -1007,6 +1046,369 @@ pappl_dnssd_unescape(
...
+        if (device && packet.object_type == _PAPPL_ASN1_OCTET_STRING && !device->uri)
+        {
+          char uri[2048];
+          snprintf(uri, sizeof(uri), "snmp://%s", (char *)packet.object_value.string.bytes);
+          device->uri = strdup(uri);
+        }
+    break;
+
+    case _PAPPL_SNMP_QUERY_DEVICE_PORT:
+        if (device && packet.object_type == _PAPPL_ASN1_OCTET_STRING)
```
michaelrsweet, line 1404: Port number is an integer in the MIBs. I guess you could check the type and convert as needed, just in case, but you definitely need to handle integers...

pappl/device.c
```diff
@@ -1007,6 +1046,373 @@ pappl_dnssd_unescape(
...
+//
+
+static int        // O - File descriptor or -1 on error
+pappl_snmp_connect(http_addr_t *addr)  // I - Device address
+{
+  int             fd = -1;      // File descriptor
+  http_addrlist_t addrlist;    // Dummy address list
+
+
+  addr->ipv4.sin_port = htons(9100);
```
michaelrsweet, line 1073: Just noticed this - it should be the port number you discovered via SNMP, so pass it in as an argument here I guess?

pappl/device.c
```diff
@@ -1007,6 +1046,373 @@ pappl_dnssd_unescape(
...
+      last_count = cupsArrayCount(devices);
+      _PAPPL_DEBUG("pappl_snmp_find: timeout=%d, last_count = %d\n", int(endtime-time(NULL)), last_count);
+    }
+  }
+
+  _PAPPL_DEBUG("pappl_snmp_find: timeout=%d, last_count = %d\n", int(endtime-time(NULL)), last_count);
+
+  for (cur_device = (_pappl_snmp_dev_t *)cupsArrayFirst(devices); cur_device; cur_device = (_pappl_snmp_dev_t *)cupsArrayNext(devices))
+  {
+    if (cur_device->port != 9100)
```
michaelrsweet, line 1174: Port number can be different than 9100. You don't want to use port 515 or 631, but anything else should be a valid raw socket port number.

pappl/device.c
```diff
@@ -1007,6 +1046,373 @@ pappl_dnssd_unescape(
...
+      continue;
+
+    if ((*cb)(cur_device->uri, cur_device->device_id, data))
+    {
+      if ((device->fd = pappl_snmp_connect(&(cur_device->address))) != -1)
```
michaelrsweet, line 1179: Pass in the port number here...

pappl/device.c
```diff
@@ -1007,6 +1046,373 @@ pappl_dnssd_unescape(
...
+
+        if (packet.object_value.oid[i] >= 0)
+        {
+          _PAPPL_DEBUG("pappl_snmp_read_response: Discarding device (not printer).\n");
+          return;
+        }
+
+        // Add the device and request the device data
+
+        temp = calloc(1, sizeof(_pappl_snmp_dev_t));
```
michaelrsweet, line 1356: Probably set the default port value to 9100 here...

Issue #10, opened 2020-04-30
Web Interface: Other issues/enhancements


michaelrsweet:
Printing Defaults is missing values for the following:

- [x] Orientation
- [x] Print Speed
- [x] Print Darkness

Would also be nice to see:

- [x] Access to log file (for file-based logging)

PR #32, opened 2020-06-26
Add PCL driver

JaiLuthra1:


michaelrsweet:
@JaiLuthra1 The sample driver isn't meant to be a part of the PAPPL library, but rather separate example code (in the testsuite directory is fine) that is used to show how to use PAPPL to create a printer application to support a PCL printer.

JaiLuthra1:
@michaelrsweet I have moved the code to the testsuite directory.

Could you please review?

testsuite/testmainloop.c
```diff
@@ -109,7 +109,7 @@ system_cb(int           num_options,	// I - Number of options
 
   papplSystemAddListeners(system, NULL);
   papplSystemSetHostname(system, hostname);
-  test_setup_drivers(system);
```
michaelrsweet, line 112: This should be a *separate* program from the testmainloop code...

testsuite/testpappl.h
```diff
@@ -24,6 +24,7 @@
 
 extern void	test_setup_drivers(pappl_system_t *system);
 extern int     testMain(int argc, char *argv[]);
+extern void papplPCLSetup(pappl_system_t *system);
```
michaelrsweet, line 27: This should be separate from the testpappl program.

testsuite/rastertohp.c
```diff
@@ -0,0 +1,878 @@
...
+// Licensed under Apache License v2.0.  See the file "LICENSE" for more
+// information.
+//
+
+//
+// Include necessary headers...
+//
+
+# include "testpappl.h"
+# include <pappl/base-private.h>
```
michaelrsweet, line 15: The example should not include private headers. The point is to provide an example of how to use PAPPL to make a printer application using just the public API headers.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+//
+
+static bool                   // O - `true` on success, `false` on failure
+pcl_status(
+    pappl_printer_t *printer) // I - Printer
+{
+  char	driver_name[256];     // Driver name
+
+
+  if (!strncmp(papplPrinterGetDriverName(printer, driver_name, sizeof(driver_name)), "deskjet-", 8))
```
michaelrsweet, line 884: Add black toner for grayscale printer.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+			*hostname,	// Hostname, if any
+			*logfile,	// Log file, if any
+			*system_name;	// System name, if any
+  pappl_loglevel_t	loglevel;	// Log level
+  int			port = 0;	// Port number, if any
+  pappl_soptions_t	soptions = PAPPL_SOPTIONS_MULTI_QUEUE | PAPPL_SOPTIONS_STANDARD | PAPPL_SOPTIONS_LOG | PAPPL_SOPTIONS_NETWORK | PAPPL_SOPTIONS_SECURITY | PAPPL_SOPTIONS_TLS;
+					// System options
+  static pappl_version_t versions[1] =	// Software versions
+  {
+    { "HP Printer App", "", "1.0 build 42", { 1, 0, 0, 42 } }
```
michaelrsweet, line 923: I would just make the version number "1.0" for this.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+  };
+
+
+  // Verify that the right callback data was sent to us...
```
michaelrsweet, line 927: This part isn't necessary for the example; we have it for the unit test in order to make sure the context pointer comes through...

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+    return (NULL);
+
+  papplSystemAddListeners(system, NULL);
+  papplSystemSetHostname(system, hostname);
+  pcl_setup(system);
+
+  papplSystemSetFooterHTML(system,
+                           "Copyright &copy; 2020 by Michael R Sweet. "
+                           "Provided under the terms of the <a href=\"https://www.apache.org/licenses/LICENSE-2.0\">Apache License 2.0</a>.");
+  papplSystemSetSaveCallback(system, (pappl_save_cb_t)papplSystemSaveState, (void *)"/tmp/hp_printer_app.state");
```
michaelrsweet, line 982: I'm thinking we want to encourage developers to store their printer application state in a more stable, per-user location... I hesitate to use something like "~/.hp_printer_app.state", since both Linux and macOS have different conventions to avoid polluting the home directory... I'll add papplSystemGet/SetStateDirectory functions (which will use the right default locations per platform/user) and then you can just pass the relative path ("hp_printer_app.state") here and in the load call below to get the desired behavior without special-casing.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+// 'pcl_setup()' - Setup PCL drivers.
+//
+
+static void
+pcl_setup(
+    pappl_system_t *system)      // I - System
+{
+  static const char * const names[] =   // Driver names
+  {
+    "hp_deskjet-150dpi-300dpi-600dpi-srgb_8",
```
michaelrsweet, line 857: I think you can just use the model names in the lists and not bother with the resolution/colorspace information.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+    unsigned          page)       // I - Page number
+{
+  unsigned            plane;    // Looping var
+  cups_page_header2_t *header = &(options->header);
+                                // Page header
+  pcl_t         *pcl = (pcl_t *)papplJobGetData(job);
+                                // Job data
+
+
+  papplLogJob(job, PAPPL_LOGLEVEL_DEBUG, "Duplex = %d\n", header->Duplex);
```
michaelrsweet, line 593: Not sure we need to log this here since PAPPL logs it already, and many of the values won't be set anyways...

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+  driver_data->rstartjob          = pcl_rstartjob;
+  driver_data->rstartpage         = pcl_rstartpage;
+  driver_data->rwrite             = pcl_rwrite;
+  driver_data->status             = pcl_status;
+  driver_data->format             = "application/vnd.hp-postscript";
+  driver_data->orient_default     = IPP_ORIENT_NONE;
+  driver_data->quality_default    = IPP_QUALITY_NORMAL;
+
+  driver_data->num_resolution = 0;
+  if (strstr(driver_name, "-150dpi"))
```
michaelrsweet, line 162: As I mention in a later comment, don't use the "-NNNdpi" and "-colorspace" convention from the PWG unit tests. Just compare model names and add all of the applicable resolutions/colorspaces.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+    driver_data->x_resolution[driver_data->num_resolution   ] = 600;
+    driver_data->y_resolution[driver_data->num_resolution ++] = 600;
+  }
+  if (driver_data->num_resolution == 0)
+  {
+    papplLog(system, PAPPL_LOGLEVEL_ERROR, "No resolution information in driver name '%s'.", driver_name);
+    return (false);
+  }
+
+  if (strstr(driver_name, "-black_1"))
```
michaelrsweet, line 185: The deskjet model does color, the rest are just grayscale...

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+  else
+  {
+    papplLog(system, PAPPL_LOGLEVEL_ERROR, "No dimension information in driver name '%s'.", driver_name);
+    return (false);
+  }
+
+  // Fill out ready and default media (default == ready media from the first source)
+  for (i = 0; i < driver_data->num_source; i ++)
+  {
+    pwg_media_t *pwg = pwgMediaForPWG(driver_data->media_ready[i].size_name);
```
michaelrsweet, line 293: I don't think you've initialized the ready media size_name fields... Here I would default the paper trays to US Letter (north america) or A4 (everywhere else) and the envelope tray to #10 (north america) or DL (everywhere else). Seems like I should add API for getting these sorts of defaults:

```
extern pwg_media_t *papplGetDefaultDocumentSize(void);
extern pwg_media_t *papplGetDefaultEnvelopeSize(void);
```

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+    driver_data->source[6]  = "envelope";
+
+    driver_data->left_right = 635;	// 1/4" left and right
+    driver_data->bottom_top = 423;	// 1/6" top and bottom
+
+    snprintf(driver_data->media_ready[0].size_name, sizeof(driver_data->media_ready[0].size_name), "na_letter_8.5x11in");
+  }
+  else if (!strncmp(driver_name, "hp_laserjet", 11))
+  {
+    snprintf(driver_data->make_and_model, sizeof(driver_data->make_and_model), "HP LaserJet");
```
michaelrsweet, line 259: Use `strlcpy(driver_data->make_and_model, "bla bla bla", sizeof(driver_data->make_and_model));` instead of snprintf.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+    driver_data->type[4] = "photographic-glossy";
+
+    driver_data->left_right = 635;	 // 1/4" left and right
+    driver_data->bottom_top = 1270;	 // 1/2" top and bottom
+
+    snprintf(driver_data->media_ready[0].size_name, sizeof(driver_data->media_ready[0].size_name), "na_letter_8.5x11in");
+  }
+  else if (!strncmp(driver_name, "hp_generic", 10))
+  {
+    snprintf(driver_data->make_and_model, sizeof(driver_data->make_and_model), "Generic PCL Laser Printer");
```
michaelrsweet, line 232: Use `strlcpy(driver_data->make_and_model, "bla bla bla", sizeof(driver_data->make_and_model));` instead of snprintf.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+  }
+  else
+  {
+    papplLog(system, PAPPL_LOGLEVEL_ERROR, "Unsupported driver name '%s'.", driver_name);
+    return (false);
+  }
+
+  if (!strncmp(driver_name, "hp_deskjet", 10))
+  {
+    snprintf(driver_data->make_and_model, sizeof(driver_data->make_and_model), "HP DeskJet");
```
michaelrsweet, line 202: Use `strlcpy(driver_data->make_and_model, "bla bla bla", sizeof(driver_data->make_and_model));` instead of snprintf.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+    {
+      papplDevicePrintf(device, "\033*b%dY", pcl->feed);
+      pcl->feed = 0;
+    }
+
+    // Write bitmap data as needed...
+
+    bytes = (header->cupsWidth + 7) / 8;
+
+    for (plane = 0; plane < pcl->num_planes; plane ++)
```
michaelrsweet, line 803: This will need some work - the B&W modes are fine since PAPPL will dither grayscale to B&W for you, but the DeskJet driver asks for grayscale or RGB so you'll need to do the dithering here...

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+  // Set graphics mode...
+
+  papplDevicePrintf(device, "\033*t%uR", header->HWResolution[0]);  // Set resolution
+
+  if (header->cupsColorSpace == CUPS_CSPACE_KCMY)
+  {
+    pcl->num_planes = 4;
+    papplDevicePuts(device, "\033*r-4U"); // Set KCMY graphics
+  }
+  else if (header->cupsColorSpace == CUPS_CSPACE_CMY)
```
michaelrsweet, line 707: You won't get CMY data from PAPPL...

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,989 @@
...
+    pappl_printer_t          *printer,	// I - Printer
+    pappl_identify_actions_t actions, 	// I - Actions to take
+    const char               *message)	// I - Message, if any
+{
+  (void)printer;
+  (void)actions;
+
+
+  if (message)
+    printf("hp-printer-app: %s\n", message);
```
michaelrsweet, line 439: Probably don't want to do anything here, just include comments about what could be done.

testsuite/hp-printer-app.c
JaiLuthra1: @michaelrsweet I looked at dithering in rastertopclx filter from cups-filters. Would that work or we need something different?

michaelrsweet: Yes that would work, as would adapting the simple ordered dither matrix in the options structure.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1044 @@
...
+    bytes = (header->cupsWidth + 7) / 8;
+
+    if (pcl->num_planes > 1)
+    {
+      dither = options->dither[y & 15];
+      line = malloc(header->cupsBytesPerLine);
+
+  	  memset(line, 0, header->cupsBytesPerLine);
+
+      for (x = 0, lineptr = line, pixptr = pixels, bit = 128, byte = 0; x < header->cupsWidth; x ++, pixptr ++)
```
michaelrsweet, line 803: In this case, you need to dither each color component (RGB) separately to produce bitmaps for CMY, and then use an AND mask of the CMY bytes to create K (black). It would also be better to allocate the line buffer in the start page function and make it part of the pcl data struct.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1044 @@
...
+      if (pcl->color_bits == 1)
+      {
+        // Send bits as-is...
+        pcl_compress_data(job, device, pcl->planes[plane], bytes, plane < (pcl->num_planes - 1) ? 'V' : 'W', header->cupsCompression);
+      }
+      else
+      {
+        // Separate low and high bit data into separate buffers.
+
+        for (count = header->cupsBytesPerLine / pcl->num_planes, plane_ptr = pcl->planes[plane], bit_ptr = pcl->bit_buffer; count > 0; count -= 2, plane_ptr += 2, bit_ptr ++)
```
michaelrsweet, line 837: This code can be removed - there is no 2-bit per component mode supported by PAPPL.

testsuite/hp-printer-app.c
JaiLuthra1: @michaelrsweet So, does this mean `pixels` has data of the form RGBRGBRGBRGB... and we need to separately dither the R,G and B parts into say, rline, bline and gline respectively and then create `line`, rather pcl->line using an AND mask of these?

michaelrsweet: Yes, RGBRGBRGB..., which can be directly dithered to CMY. Then the K plane can be generated by AND'ing the CMY planes (and you then bitwise AND the CMY planes by the ones-complement of the K plane, e.g. "CMY &= ~K").

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1071 @@
...
+  papplDevicePuts(device, "\033*r1A");  // Start graphics
+
+  if (header->cupsCompression)
+    papplDevicePrintf(device, "\033*b%uM", header->cupsCompression);  // Set compression
+
+  pcl->feed = 0; // No blank lines yet
+
+  // Allocate memory for a line of graphics...
+
+  if ((pcl->planes[0] = malloc(header->cupsBytesPerLine + pcl->num_planes)) == NULL)
```
michaelrsweet, line 735: cupsBytesPerLine is the source size. The output planes will be 1-bit per component, so you want to calculate the byte length of a line, e.g., `length = (header->cupsWidth + 7) / 8;`.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1071 @@
...
+  {
+    papplLogJob(job, PAPPL_LOGLEVEL_ERROR, "Malloc failure...\n");
+    return (false);
+  }
+
+  for (plane = 1; plane < pcl->num_planes; plane ++)
+    pcl->planes[plane] = pcl->planes[0] + plane * header->cupsBytesPerLine / pcl->num_planes;
```
michaelrsweet, line 742: and then use the same length here to calculate the offset.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1071 @@
...
+    {
+      papplDevicePrintf(device, "\033*b%dY", pcl->feed);
+      pcl->feed = 0;
+    }
+
+    // Write bitmap data as needed...
+
+    bytes = (header->cupsWidth + 7) / 8;
+
+    if (pcl->num_planes > 1)
```
michaelrsweet, line 802: OK, so you need to dither for both RGB and 8-bit gray/black - only the 1-bit black raster data can be sent directly.

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1071 @@
...
+    {
+      dither = options->dither[y & 15];
+
+      memset(pcl->line, 0, header->cupsBytesPerLine * 4);
+      memset(pcl->c, 0, header->cupsBytesPerLine);
+      memset(pcl->m, 0, header->cupsBytesPerLine);
+      memset(pcl->y, 0, header->cupsBytesPerLine);
+      memset(pcl->k, 0, header->cupsBytesPerLine);
+
+      for (x = 0, cptr = pcl->c, pixptr = pixels, bit = 128, byte = 0; x < header->cupsWidth; x ++, pixptr += 3)
```
michaelrsweet, line 812: You can dither all three colors in one loop....

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1071 @@
...
+        else
+          bit /= 2;
+      }
+
+      if (bit < 128)
+        *yptr = byte;
+
+      for (i = 0, cptr = pcl->c, mptr = pcl->m, yptr = pcl->y, kptr = pcl->k, lineptr = pcl->line; i < header->cupsBytesPerLine; i ++)
+      {
+        *kptr = *cptr & *mptr & *yptr;
```
michaelrsweet, line 868: Minor optimization - save a copy of ~K in the 'byte' variable, then you can mask using the variable. Compilers aren't always smart enough to optimize away the `~*kptr`.

Also, if you dither all three colors above, then you can generate K in the same loop as well...

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1071 @@
...
+        *cptr &= ~*kptr;
+        *mptr &= ~*kptr;
+        *yptr &= ~*kptr;
+        *lineptr ++ = *cptr ++;
+        *lineptr ++ = *mptr ++;
```
michaelrsweet, line 873: and you don't need to copy to a separate line buffer...

testsuite/hp-printer-app.c
```diff
@@ -0,0 +1,1071 @@
...
+    };
+
+    if (papplPrinterGetSupplies(printer, 0, supply) == 0)
+      papplPrinterSetSupplies(printer, (int)(sizeof(supply) / sizeof(supply[0])), supply);
+  }
+  else
+  {
+    static pappl_supply_t supply[2] =	// Supply level data
+    {
+      { PAPPL_SUPPLY_COLOR_BLACK,    "Black Ink",      true, 100, PAPPL_SUPPLY_TYPE_INK },
```
michaelrsweet, line 952: This should be toner, not ink.

PR #33, opened 2020-07-09
Added Print Test page functionality.

dsam82:


pappl/printer-webif.c
```diff
@@ -1065,6 +1068,177 @@ _papplPrinterWebSupplies(
...
+      if ((document_name = strrchr(filename, '/')) != NULL)
+        document_name ++;
+      else
+        document_name = filename;
+
+      // Get the printer URI
+      httpAssembleURI(HTTP_URI_CODING_ALL, uri, sizeof(uri), "ipp", NULL, "localhost", 0, printer->resource);
+
+      // Send a Print-Job request
+      request = ippNewRequest(IPP_OP_PRINT_JOB);
```
michaelrsweet, line 1130: You don't actually need to create an IPP request - pass NULL for the attrs argument and the defaults will be applied automatically.

pappl/printer-webif.c
```diff
@@ -1065,6 +1068,177 @@ _papplPrinterWebSupplies(
...
+      char          uri[1024],        // Printer URI
+                    filename[1024],   // Test Page filename
+                    jobfile[1024],    // Print file
+                    *document_name,   // Document name
+                    *ext,             // File extension
+                    *username;        // Username
+      int           testfd;           // File descriptor for test print file
+
+      // Determine the test page file to be used
+      if (printer->driver_data.color_supported == (PAPPL_COLOR_MODE_AUTO | PAPPL_COLOR_MODE_MONOCHROME) || printer->driver_data.color_supported == (PAPPL_COLOR_MODE_AUTO | PAPPL_COLOR_MODE_MONOCHROME | PAPPL_COLOR_MODE_AUTO_MONOCHROME))
```
michaelrsweet, line 1111: Probably look at printer->driver_data.raster_types instead, or just look for the PAPPL_COLOR_MODE_COLOR bit being set.

pappl/printer-webif.c
```diff
@@ -1065,6 +1068,177 @@ _papplPrinterWebSupplies(
...
+        else
+        {
+          job->fd = -1;
+
+          // Submit the job for processing...
+          _papplJobSubmitFile(job, jobfile);
+        }
+
+        // Redirect to printer Jobs page
+        if (papplJobGetID(job))
```
michaelrsweet, line 1207: This will always be true - better to set the status string and then only redirect if there is no status.

dsam82:
> I'd still like to see some sort of callback for doing test pages (or providing a file for printing), and obviously we need a better solution for the large PNGs.

By providing a file for printing, do you mean that the user should be given the choice to provide a png file for printing?
And how can we print large PNGs if copying is slow?

michaelrsweet:
@dsam82 I don't see allowing the user to supply a file for printing, but rather the printer application using PAPPL to register a callback that generates or references its own test file.

pappl/job.c
```diff
@@ -401,6 +392,24 @@ papplPrinterFindJob(pappl_printer_t *printer,// I - Printer
...
+//
+// '_papplJobRemoveFile()' - Remove a file in spool directory
+//
+
+void
+_papplJobRemoveFile(
+    pappl_job_t *job)
+{
+  // Only remove the file if it is in spool directory...
+  if (job->filename && !strncmp(job->filename, job->system->directory, strlen(job->system->directory)))
```
michaelrsweet, line 404: Technically you'll also want to look at job->filename[dirlen] to make sure it is '/'...

pappl/printer-webif.c
```diff
@@ -1065,6 +1068,154 @@ _papplPrinterWebSupplies(
...
+      const char    *filename = NULL; // Test Page filename
+      char          jobfile[1024],    // Print file
+                    buffer[1024],     // File Buffer
+                    *ext,             // File extension
+                    *username;        // Username
+      int           testfd;           // File descriptor for test print file
+
+      // Get the testfile to print
+      if (printer->driver_data.testfunc)
+        filename = (printer->driver_data.testfunc)(printer, buffer, sizeof(buffer));
```
michaelrsweet, line 1110: If the test function returns a file, then you create a job and submit the file.

Also, if you have a file you no longer need to copy it thanks to your changes in job.c... Just pass the original filename.

pappl/job.c
dsam82: @michaelrsweet what is `dirlen` here? Is it the length of spool directory string?

michaelrsweet: Yes, something like:

```
size_t dirlen = strlen(job->system->directory);
```

and then you can reuse it as needed.

pappl/job.c
```diff
@@ -410,9 +421,26 @@ _papplJobSubmitFile(
     pappl_job_t *job,			// I - Job
     const char  *filename)		// I - Filename
 {
+  char    *format = strrchr(filename, '.');
+
   // Save the print file information...
   job->filename = strdup(filename);
 
+  if (!strcasecmp(format, ".jpg") || !strcasecmp(format, ".jpeg"))
```
michaelrsweet, line 429: I think I'd prefer to peek at the contents of the file if the format is unknown/NULL - see the valid_doc function in pappl/ipp.c for the code. The extension can be used if the peeking doesn't yield a match.

pappl/printer-webif.c
```diff
@@ -1065,6 +1068,98 @@ _papplPrinterWebSupplies(
...
+      pappl_job_t		*job;             // New job
+      struct stat		fileinfo;	        // File information
+      const char    *filename = NULL; // Test Page filename
+      char          buffer[1024],     // File Buffer
+                    *username;        // Username
+
+      // Get the testfile to print
+      if (printer->driver_data.testfunc)
+        filename = (printer->driver_data.testfunc)(printer, buffer, sizeof(buffer));
+
```
michaelrsweet, line 1109: Wrap all of the job submission code in an "if (filename) { ... }" block...

pappl/printer-webif.c
```diff
@@ -1065,6 +1068,100 @@ _papplPrinterWebSupplies(
...
+            snprintf(path, sizeof(path), "%s/jobs", printer->uriname);
+            papplClientRespondRedirect(client, HTTP_STATUS_FOUND, path);
+            cupsFreeOptions(num_form, form);
+            return;
+          }
+        }
+      }
+      else
+        status = "Test Print File does not exist.";
```
michaelrsweet, line 1144: If the callback returns NULL, the callback already did the test page. So this should be "Test Page Printed" or something like that...

michaelrsweet: No the default is for the callback to return a file name. But if it returns NULL that means it initiated a printer self test page and no print file is needed. This is most common for thermal label printers, and I want to allow for that possibility.

PR #34, opened 2020-07-10
Add snapcraft.yaml and systemd support

JaiLuthra1:


testsuite/Makefile
```diff
@@ -41,7 +47,15 @@ depend:
 
 
 # Install everything
-install:
+install:	hp-printer-app
+	echo Installing hp printer application in $(BUILDROOT)$(bindir)...
+	$(INSTALL) -d -m 755 $(BUILDROOT)$(bindir)
+	$(INSTALL) -c -m 755 hp-printer-app $(BUILDROOT)$(bindir)
+ifeq ($(HAVE_SYSTEMD),yes)
```
michaelrsweet, line 54: Use shell commands here, they are more portable than GNU make extensions, e.g.:

```
if test "x$(unitdir)" != x; then \
    commands with ";\" at the end of them
fi
```

testsuite/Makefile
```diff
@@ -22,6 +22,12 @@ TARGETS	=	\
 #		hp-printer-app \
 
 
+HAVE_SYSTEMD := $(shell pkg-config --exists systemd 2>/dev/null && echo 'yes')
+
+ifeq ($(HAVE_SYSTEMD),yes)
+	unitdir := $(shell pkg-config --variable=systemdsystemunitdir systemd)
```
michaelrsweet, line 28: This can just be:

```
unitdir := `pkg-config --variable=systemdsystemunitdir systemd`
```

If there is no systemd, unitdir will be the empty string.

michaelrsweet, line 28: This can just be:

```
unitdir = `pkg-config --variable=systemdsystemunitdir systemd`
```

If there is no systemd, unitdir will be the empty string.

michaelrsweet, line 28: This can just be:

```
unitdir = `pkg-config --variable=systemdsystemunitdir systemd`
```

If there is no systemd, unitdir will be the empty string.

testsuite/snap/snapcraft.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+grade: stable
+confinement: strict
+
+apps:
+  hp-printer-app:
+    command: bin/hp-printer-app
+    plugs: [avahi-observe, home, log-observe, network, network-bind, network-manager, raw-usb]
+
+parts:
```
michaelrsweet, line 20: You'll need to add a part that pulls in the PAPPL source code and builds it prior to the printer application.

JaiLuthra1:
Moved to [hp-printer-app](https://github.com/michaelrsweet/hp-printer-app/pull/1)

PR #35, opened 2020-07-20
Add support for job load/save and pending job-files.

dsam82:


dsam82:
@michaelrsweet I have tried using a different approach to parse the attributes than the one in cupsFileRead/Write.
Should we add multi-valued attributes support as well as rn I have added support for singe-value attributes?

pappl/job.c
```diff
@@ -246,7 +247,8 @@ _papplJobCreate(
   ippAddInteger(job->attrs, IPP_TAG_JOB, IPP_TAG_INTEGER, "time-at-creation", (int)(job->created - printer->start_time));
 
   cupsArrayAdd(printer->all_jobs, job);
-  cupsArrayAdd(printer->active_jobs, job);
+	if (!job_id)
```
michaelrsweet, line 250: This should be based on the job-state value...

pappl/job.c
```diff
@@ -246,7 +247,8 @@ _papplJobCreate(
   ippAddInteger(job->attrs, IPP_TAG_JOB, IPP_TAG_INTEGER, "time-at-creation", (int)(job->created - printer->start_time));
```
michaelrsweet, line 247: The creation date/time needs to be preserved as well.

pappl/system-loadsave.c
```diff
@@ -18,13 +18,14 @@
 // Local functions...
 //
 
+static void parse_attrs(char *value, ipp_t **attrs);
```
michaelrsweet, line 21: I'd load and save attributes in separate files in the spool directory.

pappl/system-loadsave.c
```diff
@@ -192,6 +193,72 @@ papplSystemLoadState(
 	  sscanf(value, "%dx%ddpi", &printer->driver_data.x_default, &printer->driver_data.y_default);
 	else if (!strcasecmp(line, "sides-default"))
 	  printer->driver_data.sides_default = _papplSidesValue(value);
+  else if (!strcasecmp(line, "<Job") && value)
```
michaelrsweet, line 196: Maybe just use "Job id=nnn state=value ... name=value" syntax, and look for a .ipp file for the job to get the original attributes for a pending job?

pappl/job.c
dsam82: I am saving the attribute now in the .state file only and loading from there.

pappl/job.c
dsam82: Yes, we are adding the job to the printer active jobs array if it's still pending/processing and the job-file is still there.

pappl/job.c
```diff
@@ -53,7 +53,7 @@ papplJobCancel(pappl_job_t *job)	// I - Job
 
 
 //
-// '_papplJobCreate()' - Create a new job object.
+// '_papplJobCreate()' - Create a new/exisitng job object.
```
michaelrsweet, line 56: maybe fix the spelling in this comment? :)

pappl/job.c
```diff
@@ -120,7 +121,7 @@ _papplJobCreate(
     job->impressions = ippGetInteger(attr, 0);
 
   // Add job description attributes and add to the jobs array...
-  job->job_id = printer->next_job_id ++;
+  job->job_id = (!job_id) ? printer->next_job_id ++ : job_id;
```
michaelrsweet, line 124: I normally prefer to use the pattern "value ? value : default", so this would become `job_id ? job_id : printer->next_job_id ++`.

pappl/system-loadsave.c
```diff
@@ -192,6 +191,94 @@ papplSystemLoadState(
...
+    if ((job_value = cupsGetOption("job-processing", num_options, options)) != NULL)
+      job->processing = atol(job_value);
+    if ((job_value = cupsGetOption("job-completed", num_options, options)) != NULL)
+      job->completed = atol(job_value);
+    if ((job_value = cupsGetOption("job-impressions", num_options, options)) != NULL)
+      job->impressions = atoi(job_value);
+    if ((job_value = cupsGetOption("job-imcompleted", num_options, options)) != NULL)
+      job->impcompleted = atoi(job_value);
+
+    if ((time_at_creation = atoi(cupsGetOption("time-at-creation", num_options, options))))
```
michaelrsweet, line 237: Let's remove time-at-creation and date-time-at-creation from the job->attrs that are created and just add them in copy_job_attrs in ipp.c.

pappl/system-loadsave.c
```diff
@@ -192,6 +191,94 @@ papplSystemLoadState(
...
+
+    // Add the job to printer completed jobs array...
+    if (job->state < IPP_JSTATE_STOPPED)
+    {
+      // Load the file attributes from the spool directory...
+
+      int attr_fd;										// Attribute file descriptor
+      char job_attr_filename[256];		// Attribute filename
+
+      snprintf(job_attr_filename, sizeof(job_attr_filename), "%s/job-%d-attrs.ipp", system->directory, job->job_id);
```
michaelrsweet, line 254: This also needs the printer-id - maybe create a new _papplJobOpenFile function (copy papplJobCreateFile) to handle this for any future needs. The filename should be "/system/directory/pNNNNN-jNNNNNNNNN-job-name.ipp".

pappl/system-loadsave.c
```diff
@@ -192,6 +191,94 @@ papplSystemLoadState(
...
+      break;
+    }
+
+    if ((job = _papplJobCreate(printer, job_username, job_format, job_name, atoi(job_id), NULL)) == NULL)
+    {
+      papplLog(system, PAPPL_LOGLEVEL_ERROR, "Error creating job %s for printer %s", job_name, printer->name);
+      break;
+    }
+
+    if ((job_value = cupsGetOption("job-filename", num_options, options)) != NULL)
```
michaelrsweet, line 220: For processing efficiency, drop the "job-" prefix from all of these options...

pappl/system-loadsave.c
```diff
@@ -330,6 +418,66 @@ papplSystemSaveState(
...
+      num_options = cupsAddOption("username", job->username, num_options, &options);
+      num_options = cupsAddOption("format", job->format, num_options, &options);
+
+      if (job->filename)
+        num_options = cupsAddOption("job-filename", job->filename, num_options, &options);
+      if (job->state)
+        num_options = cupsAddIntegerOption("job-state", job->state, num_options, &options);
+      if (job->state_reasons)
+        num_options = cupsAddIntegerOption("job-state_reasons", job->state_reasons, num_options, &options);
+      if (job->created)
```
michaelrsweet, line 438: Drop the "job-" prefix since it isn't really needed in this context.

pappl/system-loadsave.c
```diff
@@ -330,6 +418,66 @@ papplSystemSaveState(
...
+        num_options = cupsAddIntegerOption("job-imcompleted", job->impcompleted, num_options, &options);
+      if ((attr = ippFindAttribute(job->attrs, "time-at-creation", IPP_TAG_INTEGER)) != NULL)
+        num_options = cupsAddIntegerOption("time-at-creation", ippGetInteger(attr, 0), num_options, &options);
+
+      if (job->attrs)
+      {
+				int attr_fd;										// Attribute file descriptor
+				char job_attr_filename[256];		// Attribute filename
+
+        snprintf(job_attr_filename, sizeof(job_attr_filename), "%s/job-%d-attrs.ipp", system->directory, job->job_id);
```
michaelrsweet, line 456: Use papplJobCreateFile for this.

michaelrsweet:
@dsam82 Sorry, I'm in crunch time with a contract right now, will try to review this later today or tomorrow...

PR #37, opened 2020-07-31
change  header and button test.html and style.css

2701gouravgoel:


michaelrsweet:
I appreciate the contribution, but what are you trying to accomplish with these changes?

2701gouravgoel:
> I appreciate the contribution, but what are you trying to accomplish with these changes?

for test.html I found there were some problems of spacing in header. Also if any user open the same tab with smaller size the header should not behave as it was before. Also to make it more nice by adding border to a button.
And I had many other documents and found it will be nice to add list type in pappl.html , also underline and highlighted hover at any link.

michaelrsweet:
Perhaps if you can attach a screenshot of the issues you are seeing and how things change? I haven't noticed any spacing issues in Safari, Chrome, or Firefox on macOS, nor on Safari for iPadOS or iOS.

As for the other style changes, I'll consider them but I want to avoid any CSS side-effects as well...

michaelrsweet:
OK, so I've looked at the formatting differences and have some feedback:

- The header changes are no good - the point is to keep the server-wide links and local links together with a separator, not to put one set of links on one side and the other set on the other side.
- Adding a border around the buttons when you hover the mouse over them causes things to move, which isn't desirable.

PR #72, opened 2020-11-13
Update autoconf

mikelolasagasti:
- Update deprecated macros
- Use AS_HELP_STRING for help strings
- Print options and flags
- Delete some generated files and update .gitignore


.gitignore
```diff
@@ -1,7 +1,7 @@
+/aclocal.m4
 /autom4te.cache
-/config.h
-/config.log
-/config.status
+/config.*
```
michaelrsweet, line 3: This will also affect config.h.in... That's why I listed the files individually.

.gitignore
```diff
@@ -1,7 +1,7 @@
...
+/configure
```
michaelrsweet, line 4: This is part of the repository (intentionally)

michaelrsweet (commented):
No.

The repository intentionally contains config.guess config.h.in because automake and autoheader cause major problems on some platforms that CUPS (and PAPPL) support but that are not well supported by the FSF.

Similarly, depending on the developer to have a compatible version of autoconf installed (where the autoconf developers have repeatedly made incompatible changes over the years) is not a recipe for success. Thus, the repository contains a "precompiled" version of the configure script.

I will consider switching to some of the newer autoconf macros (AS_HELP_STRING and AC_COMPILE_IFELSE). I'm not sure the summary at the end is necessary since autoconf already provides a blow-by-blow accounting of it and there aren't that many tests being done.

mikelolasagasti:
@michaelrsweet I closed the PR when I realized that "precompiled" was a must. I'll create a new PR for `configure.ac` with some of the changes.

Thanks for reviewing!

michaelrsweet:
@mikelolasagasti Thanks for contributing!

PR #126, opened 2021-01-15
Add support for --runstatedir configure option.

OdyX:
Debian's autoheader doesn't support partial `AC_DEFINE_UNQUOTED` or `AC_DEFINE` calls; these two need all 3 arguments.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -201,7 +201,7 @@ if test "x$PKGCONFIG" != x; then
 	if $PKGCONFIG --exists gnutls; then
 		CPPFLAGS="$CPPFLAGS `$PKGCONFIG --cflags gnutls`"
 		LIBS="$LIBS `$PKGCONFIG --libs gnutls`"
-		AC_DEFINE(HAVE_GNUTLS)
+		AC_DEFINE(HAVE_GNUTLS, 1, [Whether GnuTLS is present])
 		PKGCONFIG_GNUTLS="gnutls >= 3.0,"
 	fi
 fi
--- a/configure.ac
+++ b/configure.ac
@@ -90,7 +90,7 @@ case "`$CUPSCONFIG --api-version`" in
 esac
 
 CUPS_SERVERROOT="`$CUPSCONFIG --serverroot`"
-AC_DEFINE_UNQUOTED(CUPS_SERVERROOT, "$CUPS_SERVERROOT")
+AC_DEFINE_UNQUOTED(CUPS_SERVERROOT, "$CUPS_SERVERROOT", [CUPS Server root])
 
 
 dnl Check for pkg-config, which is used for some other tests later on...


michaelrsweet:
@OdyX Please don't run autoheader on my projects. I don't use it. I don't use automake. I don't use libtool. There is a reason - they make maintaining the build system on non-Linux systems *impossible*. They introduce dependencies on other GNU tools that I don't want. The only reason I'm even using autoconf is because there isn't a viable alternative - cmake has its own baggage and requires cmake to be installed on the build system, and PAPPL has too many dependencies to just use a simple makefile (like I do in some of my other projects).

TL;DR: Do not treat PAPPL as a GNU project. It isn't.

OdyX:
Fair enough. I don't see the point of shipping `.ac` files with broken syntax. I'll build without autoreconf, no big problem.

michaelrsweet:
@OdyX The syntax is not broken, it is just missing the optional information that autoheader depends on. I can certainly drop the configure.as file from the source tarball.

OdyX:
The problem I now run into is that Debian build tools rely on `./configure` to support autotools-standard config options such as `--runstatedir`. The ./configure as shipped doesn't support this. I'll find a way.

michaelrsweet:
@OdyX The correct solution here is to ask for --runstatedir support and not (essentially) for automake support (which opens up a rather large can of worms that I want to avoid...)

I'll re-open for that issue (although I'm not sure what I'd map --runstatedir to since the printer application doesn't have to run as root...)

michaelrsweet:
OK, looks like I just need to use the recently-released autoconf 2.70 to build the configure script. I'm fixing some warnings about deprecated macros and will push some changes (for 1.1, not going to mess with the configure script for 1.0.x) that will give you `--runstatedir`. However, at the moment the value won't be used (although I might try hooking it up to the mainloop API which chooses the state file for the printer application...)

michaelrsweet:
Tracking directory/file changes in #128.

OdyX:
Great, thank you for extracting the actionable out of my PR; and sorry if my PR (apparently the 3rd or 4th about autoconf you had closed, sorry for that) was premature.

michaelrsweet:
OK, I've made extensive changes to the configure script source, not only to resolve warnings from autoconf 2.70 but to transition most of the shell script stuff over to the `AS_xxx` macros. I also added an explicit `AC_DEFINE([PAPPL_DOES_NOT_USE_AUTOHEADER_OR_AUTOMAKE])` to cause autoheader to fail and print the message for posterity... :)

[master 5d6c0e1] One more set of updates, this time to use AS_xxx macros instead of raw shell commands.  We aren't able to cover everything (there is no AS_FOR macro) but we have covered most of the things.

Issue #157, opened 2021-03-02
Extend "ps-printer-app drivers" to also show supported device IDs


tillkamppeter:
We have discussed automated printer setup and guiding users to the correct Printer Application on the [OpenPrinting Mailing list](https://lists.linuxfoundation.org/pipermail/printing-architecture/2021/thread.html) in the thread ["Automatic printer setup with Printer Applications"](https://lists.linuxfoundation.org/pipermail/printing-architecture/2021/003934.html) and came to the conclusion that there should not be conducted an automatic setup for printers by individual Printer Applications, but instead, a central system tool should guide the user to find the Printer Application which he needs for his printer.

For this it would be useful to find out which printer models each Printer Application supports. Calling
```
$ ps-printer-app drivers
```
gives a list of all contained drivers, with machine-readable name and human-readable description:
```
generic                                 Generic PostScript Printer
brother--dcp-7025--en                   Brother DCP-7025 (en)
brother--dcp-7045-n--en                 Brother DCP-7045N (en)
brother--dcp-8020--en                   Brother DCP-8020 (en)
brother--dcp-8025-d--en                 Brother DCP-8025D (en)
brother--dcp-8040--en                   Brother DCP-8040 (en)
brother--dcp-8045-d--en                 Brother DCP-8045D (en)
[...]
```
Problem is here that one does not have information about which printer models are supported, especially as driver names and descriptions do not always contain the model names. Therefore it would be great to have also the device IDs in the list, as already suggested [here](https://lists.linuxfoundation.org/pipermail/printing-architecture/2021/003950.html).

This could be either simply added, so that the `drivers` subcommand always shows a table with three columns, the last one being the device ID, or we introduce a modifier option for the `drivers` subcommand which adds this column.

tillkamppeter:
There should also be distinction between dedicated and generic support, the former meaning model-specific support which supports all or at least most functionality of the printer (like a manufacturer-supplied PPD in the PostScript Printer Application) and the latter meaning to make the printer working via some standard property, as doing PostScript, PCL 5c/e, ... (like the "Generic PostScript printer" entry in the PostScript Printer Application).

A forth column could tell whether the support is generic or dedicated. A modifier option could also make all/generic-only/dedicated-only being shown.

michaelrsweet:
[...]

Issue #158, opened 2021-03-02
Add subcommand to simply ask whether a given printer is supported


tillkamppeter:
In issue #157 I already mentioned that we want to create a tool to guide the user to find the correct Printer Application for his printer. To facilitate this, it would be great to have a subcommand like this:
```
$ ps-printer-app is-supported SEARCH_TERM
```
where `SEARCH_TERM` can be a make/model string or device ID:
```
$ ps-printer-app is-supported 'HP OfficeJet Pro 8730'
$ ps-printer-app is-supported 'MFG:HP;MDL:HP OfficeJet Pro 8730;CMD:PCL5c,PCLXL,POSTSCRIPT,NATIVEOFFICE,PDF,PCL3GUI,PCL3,PJL,Automatic,JPEG,PCLM,AppleRaster,PWGRaster,802.11,802.3,DESKJET,DYN;CLS:PRINTER;DES:D9L20A;'
```
One could call the auto-add callback function and if it returns a driver name, return this driver name and exit with 0, if it does not return a driver name (unsupported printer) return nothing and exit with 1.

This function should not run device discovery nor poll information from a device, so that it returns quickly and is independent of whether the device actually exists and whether it is connected.

tillkamppeter:
There should also be distinction between dedicated and generic support, the former meaning model-specific support which supports all or at least most functionality of the printer (like a manufacturer-supplied PPD in the PostScript Printer Application) and the latter meaning to make the printer working via some standard property, as doing PostScript, PCL 5c/e, ... (like the "Generic PostScript printer" entry in the PostScript Printer Application).

There could be different exit code, 0 for dedicated, 1 for generic, and 2 for no support.

michaelrsweet:
@tillkamppeter Using exit codes isn't viable for a number of reasons. Returning the driver name and the 1284 device ID that matches is probably the best approach and allows the caller to determine how good the match is (and would match how "drivers" will work...)

PR #168, opened 2021-06-26
Extending drivers subcommand to also show supported device IDs

Bhavna2020:
this fixes issue #157.

### DIFF
--- a/pappl/mainloop-subcommands.c
+++ b/pappl/mainloop-subcommands.c
@@ -781,7 +781,10 @@ _papplMainloopShowDrivers(
   }
 
   for (i = 0; i < system->num_drivers; i ++)
-    printf("%-39s %s\n", system->drivers[i].name, system->drivers[i].description);
+	{
+		printf("%-39s %-39s %s\n", system->drivers[i].name, system->drivers[i].description,
+		system->drivers[i].device_id);
+	}
 
   papplSystemDelete(system);
 


michaelrsweet (changes requested):
OK, as the primary consumer of this data is going to be system-level tools for auto-adding printers, I think we need to make the output more software-friendly.

The current output has the driver name (no spaces allowed) followed by whitespace and then the description (anything allowed besides newlines). At the very least we should add quotes around the description and device ID strings, e.g.:

    driver_name "Driver Description" "MFG:Foo;MDL:Bar;CMD:PCL;..."

We could also use CSV or some other well-known format to make it easier to parse. Check with @tillkamppeter to see what his preferred format will be.

tillkamppeter:
@michaelrsweet thanks for the hint, I gace a similar hint on issue #158 but not on issue #157.
The use of quotes for column 2 and 3 is OK here, a CSV would be not such a good idea as the device ID can contain commas.
@Bhavna2020, could you implement it this way? Thanks.

Bhavna2020:
OK, I have made the changes.

tillkamppeter:
@Bhavna2020, to check whether a Printer Application supports a printer specified by its device ID you **should not** parse ythe ID and go through the driver list by yourself but instead, let the Printer Application do exactly the same as it does when you add the printer with driver selection "auto". You need to call the auto-add callback function with the device ID of the printer as argument (`system->autoadd_cb`).

PR #169, opened 2021-06-28
Extend drivers subcommand to also show supported device IDs

Bhavna2020:


pappl/mainloop-subcommands.c
```diff
@@ -781,8 +781,9 @@ _papplMainloopShowDrivers(
   }
 
   for (i = 0; i < system->num_drivers; i ++)
-    printf("%-39s %s\n", system->drivers[i].name, system->drivers[i].description);
-
+  {
+		printf("%-39s \"%s%-30s \"%s\" \n", system->drivers[i].name, system->drivers[i].description, "\"" , system->drivers[i].device_id);
```
michaelrsweet, line 785: Not sure what the point of the fixed width formatting is for? Should be enough to put the description and device ID in quotes.

michaelrsweet (changes requested):

Bhavna2020:
OK.

PR #171, opened 2021-07-03
Added subcommand to simply ask whether a given printer is supported

Bhavna2020:
This fixes issue #158.

### DIFF
--- a/pappl/mainloop-subcommands.c
+++ b/pappl/mainloop-subcommands.c
@@ -137,7 +137,7 @@ _papplMainloopCheckPrinterSupport(
 {
   const char           *driver_name;	// I - Driver name
   const char           *device_id;	// I - IEEE-1284 device ID
-  int			i;								// Looping variable
+  int			i;		// Looping variable
   pappl_system_t	*system;	// System object
 						
   device_id = cupsGetOption("device-id", num_options, options);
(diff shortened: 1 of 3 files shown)


pappl/mainloop-subcommands.c
```diff
@@ -122,6 +122,59 @@ _papplMainloopAddPrinter(
...
+						
+  device_id = cupsGetOption("device-id", num_options, options);
+  
+  if (!device_id)
+  {
+   	fprintf(stderr, "%s: Missing '-o device-id=DEVICE-ID'.\n", base_name);
+    return(1);
+  }
+  
+  if (!system_cb)
```
michaelrsweet, line 151: The system callback doesn't need to be set - look at the code for the drivers sub-command...

pappl/mainloop.c
```diff
@@ -84,6 +84,7 @@ papplMainloop(
   {					// List of standard sub-commands
     "add",
     "autoadd",
+    "is-supporting",
```
michaelrsweet, line 87: I would prefer to see this become part of the "drivers" sub-command.

michaelrsweet (changes requested):
Please add support for the "device-id" option to the "drivers" sub-command.

Bhavna2020:
OK.

PR #175, opened 2021-07-08
Add scanner.c and scanner-accessors.c files

Bhavna2020:


michaelrsweet (changes requested):
You'll need to update the scanner attributes to those listed in the [IPP Scan Service specification](https://ftp.pwg.org/pub/pwg/candidates/cs-ippscan10-20140918-5100.17.pdf).

michaelrsweet (changes requested):
There is still a lot of print-specific stuff in here - please read the IPP Scan Service spec.

Also, we need to support pairing scanners with printers, since the typical use case (multi-function printer) will have the scan-specific TXT keys added to the printer, and the printer-dns-sd-name value coming from the printer.

PR #176, opened 2021-07-15
Add scanner-driver.c

Bhavna2020:


michaelrsweet (changes requested):
Please fix the attribute names.

Also, you need to remove print-specific content and add scan-specific content.

PR #192, opened 2022-02-08
Configure script: Use better papplstatedir and papplsockdir defaults

GravisZro:
The --with-papplstatedir and --with-papplsockdir seem to be duplicative of --localstatedir and --runstatedir.  As such the default values should be the values of --localstatedir and --runstatedir.  Perhaps they should be removed in the future.


michaelrsweet (changes requested):
Unfortunately, it is not enough to just substitute "runstatedir" and "localstatedir" since configure will happily just substitute "${prefix}/var" and "${prefix}/var/run" in config.h.

Also, the changes need to be made to "configure.ac"...

PR #223, opened 2022-10-19
adding basic printer description attributes

wifiprintguy:
using a patchfile from a diff from the epx-prototype branch on this one file because I cannot figure out the cherry-pick syntax and am doing it the hard way


wifiprintguy:
Found a blog with a tutorial on how to retroactively sign Git commits so did that here

pappl/printer.c
```diff
@@ -174,6 +175,80 @@ papplPrinterCreate(
     "gzip",
     "none"
   };
+
```
michaelrsweet, line 178: I'd prefer not to have "EPX 2.0" comments here with the extra separating whitespace. Also, the current implementation of the cupsHashData function in libcups only supports the `sha2-` algorithms and not the `sha3-` or `shake-` algorithms, so they shouldn't be listed here. Finally, indentation should be *2* spaces, not 4.

pappl/printer.c
```diff
@@ -630,6 +705,62 @@ papplPrinterCreate(
   // which-jobs-supported
   ippAddStrings(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "which-jobs-supported", sizeof(which_jobs) / sizeof(which_jobs[0]), NULL, which_jobs);
 
+
+    //-----------------------------------------------------------------------------------
```
michaelrsweet, line 709: Ditto here about EPX comments and extra whitespace.

Additionally, I do try to sort the attributes since it helps to speed up filtering with "requested-attributes", so please move these around to appear inline with the other existing attributes.

pappl/printer.c
```diff
@@ -630,6 +705,62 @@ papplPrinterCreate(
...
+    ippAddStrings(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "job-storage-disposition-supported", (int)(sizeof(job_storage_disposition) / sizeof(job_storage_disposition[0])), NULL, job_storage_disposition);
+
+    // job-storage-group-supported - EPX 2.0
+    ippAddStrings(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "job-storage-group-supported", (int)(sizeof(job_storage_group) / sizeof(job_storage_group[0])), NULL, job_storage_group);
+
+    // job-storage-supported - EPX 2.0
+    ippAddStrings(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "job-storage-supported", (int)(sizeof(job_storage) / sizeof(job_storage[0])), NULL, job_storage);
+
+    // printer-detailed-status-messages  - EPX 2.0
+    ippAddString(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_TEXT), "printer-detailed-status-messages", NULL, "TODO: HOOK ME UP TO SOMETHING"); // TODO: Hook this up to something that reports the detailed status
```
michaelrsweet, line 753: Since this is Printer Status attribute, this should be in `_papplPrinterCopyAttributes`. And honestly there is nothing we'd be filling in here except maybe a history buffer of prior printer errors (i.e. capture `papplLogPrinter` errors).

michaelrsweet (changes requested):
Good first pass, mostly editorial changes needed.

wifiprintguy:
I think I resolved all the issues with 2 additional commits. let me know if I need to do more.

pappl/printer.c
```diff
@@ -503,24 +556,63 @@ papplPrinterCreate(
...
   ippAddRange(printer->attrs, IPP_TAG_PRINTER, "job-k-octets-supported", 0, k_supported);
 
+  // job-password-encryption-supported
+  ippAddStrings(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "job-password-encryption-supported", (int)(sizeof(job_password_encryption) / sizeof(job_password_encryption[0])), NULL, job_password_encryption);
+
+  // job-password-length-supported
+  ippAddRange(printer->attrs, IPP_TAG_PRINTER, "job-password-length-supported", 4, 1020); // 4 digit PIN up to 255 characters UTF-8 with 4 octets / character
+
+  // job-password-repertoire-configured
+  ippAddString(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "job-password-repertoire-configured", NULL, "iana_utf-8_any");
```
michaelrsweet, line 578: This should be a settable member of `pappl_printer_t` that is reported in `_papplPrinterCopyAttributes`, and should internally be an enum that is mapped to the strings.

pappl/printer.c
```diff
@@ -503,24 +556,63 @@ papplPrinterCreate(
...
+  ippAddInteger(printer->attrs, IPP_TAG_PRINTER, IPP_TAG_INTEGER, "job-password-supported", 255);
+
   // job-priority-default
   ippAddInteger(printer->attrs, IPP_TAG_PRINTER, IPP_TAG_INTEGER, "job-priority-default", 50);
 
   // job-priority-supported
   ippAddInteger(printer->attrs, IPP_TAG_PRINTER, IPP_TAG_INTEGER, "job-priority-supported", 1);
 
+  // job-release-action-default
+  ippAddString(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "job-release-action-default", NULL, "none");
```
michaelrsweet, line 593: The default needs to be part of `pappl_printer_t` and reported in `_papplPrinterCopyAttributes`. Also, I've been using enums at the API level and mapping to strings at the last minute...

pappl/printer.c
```diff
@@ -616,6 +708,9 @@ papplPrinterCreate(
   // printer-uuid
   ippAddString(printer->attrs, IPP_TAG_PRINTER, IPP_TAG_URI, "printer-uuid", NULL, uuid);
 
+  // proof-copies-supported
```
michaelrsweet, line 711: Looking more closely at the existing code, I think we need to make this conditional based on copy support - see how the "copies-supported" attribute is handled in `_papplPrinterCopyAttributes`.

michaelrsweet (changes requested):
Better, but still some nits...

pappl/printer.c
```diff
@@ -174,6 +174,59 @@ papplPrinterCreate(
...
+    "iana_us-ascii_letters",
+    "iana_us-ascii_complex",
+    "iana_us-ascii_any",
+    "iana_utf-8_digits",
+    "iana_utf-8_letters",
+    "iana_utf-8_any"
+  };
+  static const char * const job_release_action[] =
+  {                    // job-release-action-supported values
+    "none",
```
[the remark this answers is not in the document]
wifiprintguy: I was thinking it was necessary to do so to satisfy the prototyping requirement for the EPX spec.

pappl/printer.c
```diff
@@ -503,24 +556,63 @@ papplPrinterCreate(
   // ipp-versions-supported
   ippAddStrings(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "ipp-versions-supported", (int)(sizeof(ipp_versions) / sizeof(ipp_versions[0])), NULL, ipp_versions);
 
+  // job-cancel-after-default
+  ippAddInteger(printer->attrs, IPP_TAG_PRINTER, IPP_TAG_INTEGER, "job-cancel-after-default", 900); // 15 minutes
```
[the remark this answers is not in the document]
wifiprintguy: should the value be set from a new field in _pappl_printer_s, like "cancel_after"?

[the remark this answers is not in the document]
michaelrsweet: Yes

pappl/printer.c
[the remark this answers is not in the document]
michaelrsweet: I think it is sufficient to show that it works for one, since the spec doesn't require support for all actions, and the attribute itself is only conditionally required (condition not specified, but I assume for Printers that support Job Release).

From an implementation standpoint, any of these can be supported through the web interface. But longer term I'd want to hook this up to the printer driver data so the driver can indicate which methods are natively supported by the printer.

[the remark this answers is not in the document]
wifiprintguy: I will start with 'button-press' and add the others later if I get a bee in my bonnet

pappl/printer.c
```diff
@@ -503,24 +556,63 @@ papplPrinterCreate(
...
+  ippAddRange(printer->attrs, IPP_TAG_PRINTER, "job-cancel-after-supported", 30, 14400); // 30 seconds to 4 hours
+
   // job-ids-supported
   ippAddBoolean(printer->attrs, IPP_TAG_PRINTER, "job-ids-supported", 1);
 
   // job-k-octets-supported
   ippAddRange(printer->attrs, IPP_TAG_PRINTER, "job-k-octets-supported", 0, k_supported);
 
+  // job-password-encryption-supported
+  ippAddStrings(printer->attrs, IPP_TAG_PRINTER, IPP_CONST_TAG(IPP_TAG_KEYWORD), "job-password-encryption-supported", (int)(sizeof(job_password_encryption) / sizeof(job_password_encryption[0])), NULL, job_password_encryption);
```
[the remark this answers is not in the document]
wifiprintguy: Presumably this is also true for the line for "job-password-repertoire-supported" and "job-release-action-supported"?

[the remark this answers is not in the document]
michaelrsweet: Yes

pappl/printer.c
wifiprintguy: These enums can be consecutive integer values, not bits for a bitfield, so I'll add them that way. Do you want a "none" value mapped to 0 (zero)?

michaelrsweet: Not all values will be supported by all printers, so we should use the bitfield approach like I do for print-color-mode and other keyword attributes. There are support methods to do the mapping...

pappl/printer.c
wifiprintguy: Maybe "job-sheets-default" needs to similarly be moved into _papplPrinterCopyAttributes()?

michaelrsweet: Since we don't currently support job-sheets in PAPPL and the only value we report is "none", I don't think we need/want to do that at this time.

Obviously if we add job-sheets/-col support to PAPPL then the default will need to move to `_papplPrinterCopyAttributes` like the others.

PR #234, opened 2022-11-04
Generate a device ID if expected fields are missing

abdelq:
Some devices are not setting the device ID properly. In the case of the cheap printer I have, it's just whitespace. Without this change I end up with `usb://Unknown/Unknown?serial=Printer`. Not very useful for lprint's `autoadd` for example...

### DIFF
--- a/pappl/device-usb.c
+++ b/pappl/device-usb.c
@@ -362,12 +362,12 @@ pappl_usb_find(
             if (device->handle)
             {
               // Build the device URI...
-              const char *make,		// Pointer to make
-			*model,		// Pointer to model
-			*serial = NULL;	// Pointer to serial number
-	      char	*ptr,		// Pointer into device ID
-			copy_did[1024],	// Copy of device ID
-			temp[256];	// Temporary string for serial #
+              char	*make,			// Pointer to make
+			*model,			// Pointer to model
+			*serial;		// Pointer to serial number
+	      char	tmp[256],		// Temporary string for serial #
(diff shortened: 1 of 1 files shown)


michaelrsweet (changes requested):
I think I want to generate a device ID string in this case, but want to dig a bit more to be sure. But in concept I am OK with this change.

abdelq:
Alright, second attempt. Feedback appreciated.

michaelrsweet:
OK, I've pushed a similar change:

[master 1a28cf9] Generate device ID as needed (Issue #234)

Can you review and provide feedback?

michaelrsweet (changes requested):
Pushed a slightly different (but similar) change - basically always get the serial, manufacturer, and product strings and cache the device ID. Let me know what you think!

abdelq:
Thanks Michael. This is more in line with the alternative implementation I was thinking of. My only comment is that setting `'\0'` with `libusb_get_string_descriptor_ascii` is redundant.

Refs:
- https://github.com/libusb/libusb/blob/f3619c4078a921c3d6bf3238ff7a8dc70d60e40e/libusb/descriptor.c#L1127
- https://github.com/libusb/libusb/blob/f3619c4078a921c3d6bf3238ff7a8dc70d60e40e/libusb/descriptor.c#L1137

michaelrsweet:
Noted, and I pushed a change to simplify:

[master 84279c3] Clean up libusb API usage (Issue #234)

PR #236, opened 2022-11-09
Rename papplDeviceParse1284ID to papplDeviceParseID in manpages

abdelq:


michaelrsweet:
Aside from breaking binary compatibility (which would push this to a 2.0 change at the very least), it is named as it is because the device ID comes from the IEEE 1284 standard. Given that we have Job IDs, Printer IDs, etc. that are simple integers, "ParseID" is too non-specific so I want to keep the name.

PR #246, opened 2023-01-06
EPX Basic Support - tranche 1

wifiprintguy:
Made a new branch from the upstream michaelrsweet/pappl libcups3 branch, diffed against and manually merged from my epx-proto-libcups3 branch to drop all my whitespace changes that were not helpful to the effort in the slightest. What is here is basic support for EPX Job Storage and for other features. Still some work to be done in the following areas:
* Job Storage needs a web page that can present the jobs for a given user or entity, etc.
* Job Proof and Suspend needs the Job to be put into the 'processing-stopped' state after the proof copies have been produced
* Job Print Policy needs to be implemented
* Job Storage Groups needs to be implemented


wifiprintguy:
Withdrawn - additional work needed

Issue #253, opened 2023-02-09
Print queue setup fails. URI rejected.


debiantriage:
**Describe the bug**

I submitted this issue at https://github.com/OpenPrinting/pappl-retrofit/issues/8 first. Basically, installation of a print queue within gutenprint-printer-app fails.

**Steps to reproduce the behavior:**
1. Find URI:
``root@test-new:~# gutenprint-printer-app devices
usb://EPSON/Stylus%20Photo%20RX420?serial=LJ4O40407152301540``
2. Establish queue:
``gutenprint-printer-app add -d RX420 -m epson--stylus-photo-rx-420--en -v usb://EPSON...``

**Outcome**
``gutenprint-printer-app: Unable to add printer: Unsupported smi55357-device-uri uri value.``

**Resolution**
Use ``cups:usb://EPSON...`` for the URI.``

**Expected behavior**
Establish a queue as documented.

**System Information:**
 - Debian unstale.
 - gutenprint-printer-app from the snap store.

michaelrsweet:
Hmm, so a couple things to try:

1. Make sure to put quotes around the device URI ('%' and '?' can be interpreted by the shell)
2. Check the gutenprint-printer-app log file/syslog for details of what URI was seen

debiantriage:
> Hmm, so a couple things to try:
> 
>     1. Make sure to put quotes around the device URI ('%' and '?' can be interpreted by the shell)

Tried that, without success

> 
>     2. Check the gutenprint-printer-app log file/syslog for details of what URI was seen

The relevant lines from the log in /var/snap appear to be
```
D [2023-02-09T14:00:38.141Z] [Client 8] Create-Printer request:   smi55357-device-uri uri usb://EPSON/Stylus%20Photo%20RX420?serial=LJ4O40407152301540
```
```
I [2023-02-09T14:00:38.141Z] [Client 8] Create-Printer client-error-attributes-or-values-not-supported (Unsupported smi55357-device-uri uri value.)
```
```
D [2023-02-09T14:00:38.141Z] [Client 8] Create-Printer response:   status-message textWithoutLanguage Unsupported smi55357-device-uri uri value.
```

Anyway, I get the same outcome with  ps-printer-app and a networked printer. Log attached.

[psq.log.txt](https://github.com/michaelrsweet/pappl/files/10698043/psq.log.txt)

michaelrsweet:
OK, so I just can't get it to fail with the test suite. What version of PAPPL are you using?

[...]

PR #262, opened 2023-03-05
Fix listing devices via "devices" subcommand

tillkamppeter:
Fixes issue #253

When running a Printer Application (ex. ps-printer-app) as server and doing the client call
```
ps-printer-app devices
```
the use of custom backends (schemes, like "`cups:`" for CUPS backends in pappl-retrofit) was not taken into account, as simply `papplDeviceList()` was called, with neither a system created nor the devices list polled from the running server.

One could create the system before to get the custom scheme definitions in place, but the server is usually running as root and the client doing the request as the calling user. As some backends, independent of being PAPPL's own backends or any custom backends, discover devices only as root, letting the client call producing a different usually incomplete result.

Therefore we do an `IPP_OP_PAPPL_FIND_DEVICES` IPP request to the server to let the server auto-discover the available devices and report back the results.

There is also a bug in the server's `IPP_OP_PAPPL_FIND_DEVICES` (and `IPP_OP_PAPPL_CREATE_PRINTERS`) operation. The device list callback function `_papplDeviceInfoCallback()` returns "`true`" and not "`false`" after outputting an item, making the device listing process stop on the first item. making only the first discovered device listed (or only the first discovered printer set up). This is also corrected in this commit.

### DIFF
--- a/pappl/mainloop-subcommands.c
+++ b/pappl/mainloop-subcommands.c
@@ -811,42 +811,37 @@ _papplMainloopShowDevices(
   response = cupsDoRequest(http, request, "/ipp/system");
   httpClose(http);
 
-  if (cupsLastError() != IPP_STATUS_OK)
+  if (cupsLastError() != IPP_STATUS_OK && cupsLastError() != IPP_STATUS_ERROR_NOT_FOUND)
   {
     _papplLocPrintf(stderr, _PAPPL_LOC("%s: Unable to get available devices: %s"), base_name, cupsLastErrorString());
+    if (response)
+      ippDelete(response);
     return (1);
   }
 
   if ((attr = ippFindAttribute(response, "smi55357-device-col", IPP_TAG_BEGIN_COLLECTION)) != NULL)
   {
-    int num_devices = ippGetCount(attr);
-    for (int i = 0; i < num_devices; i ++)
-    {
-      ipp_t* item = ippGetCollection(attr, i);
-      ipp_attribute_t *item_attr;
-      const char *str;
+    cups_len_t	i,				// Looping var
+		num_devices = ippGetCount(attr);// Number of device entries
 
-      if ((item_attr = ippFindAttribute(item, "smi55357-device-uri", IPP_TAG_ZERO)) != NULL &&
-	  (str = ippGetString(item_attr, 0, NULL)) != NULL)
+    for (i = 0; i < num_devices; i ++)
+    {
+      ipp_t		*item = ippGetCollection(attr, i);	// Device entry
+      ipp_attribute_t	*item_attr;				// Device attr
+
+      if ((item_attr = ippFindAttribute(item, "smi55357-device-uri", IPP_TAG_ZERO)) != NULL)
       {
-	puts(str);
+	puts(ippGetString(item_attr, 0, NULL));
 
 	if (cupsGetOption("verbose", num_options, options))
 	{
-	  if ((item_attr = ippFindAttribute(item, "smi55357-device-info", IPP_TAG_ZERO)) != NULL &&
-	      (str = ippGetString(item_attr, 0, NULL)) != NULL)
-	    printf("    %s\n", str);
-	  if ((item_attr = ippFindAttribute(item, "smi55357-device-id", IPP_TAG_ZERO)) != NULL &&
-	      (str = ippGetString(item_attr, 0, NULL)) != NULL)
-	    printf("    %s\n", str);
+	  if ((item_attr = ippFindAttribute(item, "smi55357-device-info", IPP_TAG_ZERO)) != NULL)
+	    printf("    %s\n", ippGetString(item_attr, 0, NULL));
+	  if ((item_attr = ippFindAttribute(item, "smi55357-device-id", IPP_TAG_ZERO)) != NULL)
+	    printf("    %s\n", ippGetString(item_attr, 0, NULL));
 	}
       }
     }
-  }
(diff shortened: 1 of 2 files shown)


pappl/mainloop-subcommands.c
```diff
@@ -797,7 +795,61 @@ _papplMainloopShowDevices(
...
+
+  if (cupsLastError() != IPP_STATUS_OK)
+  {
+    _papplLocPrintf(stderr, _PAPPL_LOC("%s: Unable to get available devices: %s"), base_name, cupsLastErrorString());
+    return (1);
+  }
+
+  if ((attr = ippFindAttribute(response, "smi55357-device-col", IPP_TAG_BEGIN_COLLECTION)) != NULL)
+  {
+    int num_devices = ippGetCount(attr);
```
michaelrsweet, line 822: Needs to be `cups_len_t` since CUPS 3.x uses `size_t` while CUPS 2.x uses `int`. Also please add a comment describing the variable...

pappl/mainloop-subcommands.c
```diff
@@ -797,7 +795,61 @@ _papplMainloopShowDevices(
...
+    for (int i = 0; i < num_devices; i ++)
```
michaelrsweet, line 823: Prefer to declare loop variable outside the `for` loop (yes I know C99 allows it but then you don't get to document what "i" is...) Also it needs to be `cups_len_t`...

pappl/mainloop-subcommands.c
```diff
@@ -797,7 +795,61 @@ _papplMainloopShowDevices(
...
+    {
+      ipp_t* item = ippGetCollection(attr, i);
```
michaelrsweet, line 825: Comments for each variable, should be `ipp_t *item` (space before "*" and not after).

pappl/mainloop-subcommands.c
```diff
@@ -797,7 +795,61 @@ _papplMainloopShowDevices(
...
+      ipp_attribute_t *item_attr;
```
michaelrsweet, line 826: Also comments for these plus use tabs to align "*name".

"str" might be better "device_uri" (more descriptive)...

pappl/mainloop-subcommands.c
```diff
@@ -797,7 +795,61 @@ _papplMainloopShowDevices(
...
+      const char *str;
+
+      if ((item_attr = ippFindAttribute(item, "smi55357-device-uri", IPP_TAG_ZERO)) != NULL &&
```
michaelrsweet, line 829: This can all be on one line, and "str" cannot be `NULL` so maybe don't even worry about assigning to a variable and just `puts(ippGetString(item_attr, 0, NULL);`...

pappl/mainloop-subcommands.c
```diff
@@ -797,7 +795,61 @@ _papplMainloopShowDevices(
...
+	  (str = ippGetString(item_attr, 0, NULL)) != NULL)
+      {
+	puts(str);
+
+	if (cupsGetOption("verbose", num_options, options))
+	{
+	  if ((item_attr = ippFindAttribute(item, "smi55357-device-info", IPP_TAG_ZERO)) != NULL &&
+	      (str = ippGetString(item_attr, 0, NULL)) != NULL)
```
michaelrsweet, line 837: Again, the string cannot be `NULL` if the attribute isn't `NULL`. Another common code pattern I use is `if ((variable = ippGetString(ippFindAttribute(item, "bla", IPP_TAG_ZERO), 0, NULL)) != NULL)`.

pappl/mainloop-subcommands.c
```diff
@@ -797,7 +795,61 @@ _papplMainloopShowDevices(
...
+	    printf("    %s\n", str);
+	  if ((item_attr = ippFindAttribute(item, "smi55357-device-id", IPP_TAG_ZERO)) != NULL &&
+	      (str = ippGetString(item_attr, 0, NULL)) != NULL)
+	    printf("    %s\n", str);
+	}
+      }
+    }
+  }
+  else if (cupsLastError() != IPP_STATUS_ERROR_NOT_FOUND)
```
michaelrsweet, line 846: You already checked `cupsLastError` before, so if there is no attribute there are no devices (and PAPPL will return an IPP error which you handle above...)

michaelrsweet (changes requested):
Generally I like this, just needs some cleanup.

pappl/mainloop-subcommands.c
tillkamppeter: Done.

pappl/mainloop-subcommands.c
tillkamppeter: Done.

pappl/mainloop-subcommands.c
tillkamppeter: Done (`str` removed, see below).

pappl/mainloop-subcommands.c
tillkamppeter: Done.

pappl/mainloop-subcommands.c
tillkamppeter: Done.

pappl/mainloop-subcommands.c
tillkamppeter: Done as in your previous comment.

pappl/mainloop-subcommands.c
tillkamppeter: Done.

michaelrsweet (approved):

PR #272, opened 2023-04-12
log.c: Fix crash due negative buffer size

zdohnal:
A higher pointer address is subtracted from lower pointer address in `write_log`, which causes crash if pappl is built with _FORTIFY_SOURCE=3.

It can be reproduced by starting `lprint` service.

### DIFF
--- a/pappl/log.c
+++ b/pappl/log.c
@@ -565,7 +565,7 @@ write_log(pappl_system_t   *system,	// I - System
 	case 'e' :
 	case 'f' :
 	case 'g' :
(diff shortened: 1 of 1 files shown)


zdohnal:
Ok, I didn't know Github project can be set to require signing - learnt something new :) .

michaelrsweet:
[master 9637431] Fix a potential buffer overflow in the logging code (Issue #272)

[v1.3.x 0c69b3d] Mirror fix from master.

zdohnal:
Thanks, @michaelrsweet !

PR #276, opened 2023-06-12
Fixes for Avahi-based DNS-SD registration

tillkamppeter:
- If we provide the service on localhost only, we have to register with the interface index of the loopback device "lo". This is NOT 0. the Interface index 0 is reserved for if_nametoindex() function to report an error condition. So we have to use if_nametoindex("lo") to get the actual index (which is probably not always the sane, as its actual value is nowhere documented).

- When we call any avahi_entry_group_add_...() function, the second argument is the interface index, and we have to use the one we selected in the beginning (if_index). The forth parameter is a for a set of flags, which one should set to 0 ("flags: Only != 0 if you really know what you do"). It is for sure wrong to supply the interface index here.

### DIFF
--- a/pappl/dnssd-private.h
+++ b/pappl/dnssd-private.h
@@ -20,6 +20,7 @@
 #    include <avahi-common/domain.h>
 #    include <avahi-common/error.h>
 #    include <avahi-common/thread-watch.h>
+#    include <net/if.h>
 #  endif // HAVE_MDNSRESPONDER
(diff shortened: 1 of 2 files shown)


michaelrsweet (approved):
+1

michaelrsweet:
[v1.3.x bef9c2e] Mirror Avahi fix (Issue #276)

PR #318, opened 2024-01-10
configure.ac: Fix --enable-libpam option

zdohnal:
Previously the option was reported as unknown when used, because the feature name in `configure.ac` was only `pam` and `autoconf` defines available options from its name, but `./configure -h` showed `--enable-libpam`.

I fixed the option to follow the help message, in case someone was using it already.

### DIFF
--- a/configure
+++ b/configure
@@ -737,7 +737,7 @@ enable_option_checking
 enable_libjpeg
 enable_libpng
 enable_libusb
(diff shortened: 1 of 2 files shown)


michaelrsweet (approved):
+1

Funny, I probably never noticed this because the default is to use PAM if available.

PR #319, opened 2024-01-11
device-network.c: Fix crash in `pappl_dnssd_list()` with Avahi

zdohnal:
If PAPPL testsuite runs in isolated environment, e.g. in mock, we cannot generate Avahi client, because Avahi is not running. This causes `_papplDNSSDInit()` to return NULL, which is later sent into `avahi_service_browser_new()`, which uses `assert()` for this argument and make the binary crash if the input is NULL.

The crash makes the testsuite fail during the build, which is useful sanity check before the package is built.

### DIFF
--- a/pappl/device-network.c
+++ b/pappl/device-network.c
@@ -416,7 +416,10 @@ pappl_dnssd_list(
 
 
   devices.types   = types;
-  devices.dnssd   = cupsDNSSDNew(err_cb, err_data);
+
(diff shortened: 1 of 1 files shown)


michaelrsweet (approved):

michaelrsweet:
This shouldn't cause a crash because the cupsDNSSD APIs are sane about NULL checking/ignoring. But happy to add this to short-circuit things...

PR #322, opened 2024-01-19
printer-driver.c: Fix crash when adding generic PS printer

zdohnal:
If legacy-printer-app is run for the first time and generic PS driver was chosen when installing the printer, the application crashes during adding the printer.

It is because the printer structure does not have device_id set, and the library tries to get data from already freed `printer->driver_attrs`. Additionally, if the device id exists in the printer, it rewrites incorrect structure (we should set device id attr in `attrs`, which we return).

The fix adds possibility to extract device id from `printer->attrs` and if device id is missing in `printer->attrs` and `document-format-supported` as well, try `document-format-default` and if this attribute is missing, return NULL.

### DIFF
--- a/pappl/printer-driver.c
+++ b/pappl/printer-driver.c
@@ -968,7 +968,11 @@ make_attrs(
   // printer-device-id
   if (printer->device_id)
   {
-    ippAddString(printer->attrs, IPP_TAG_PRINTER, IPP_TAG_TEXT, "printer-device-id", NULL, printer->device_id);
+    ippAddString(attrs, IPP_TAG_PRINTER, IPP_TAG_TEXT, "printer-device-id", NULL, printer->device_id);
+  }
+  else if ((attr = ippFindAttribute(printer->attrs, "printer-device-id", IPP_TAG_PRINTER)) != NULL)
+  {
+    ippAddString(attrs, IPP_TAG_PRINTER, IPP_TAG_TEXT, "printer-device-id", NULL, ippGetString(attr, 0, NULL));
   }
   else
   {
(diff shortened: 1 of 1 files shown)


zdohnal:
The key for reproducing the issue is that it must be the first run of legacy-printer-app after its installation and start. I was not able to reproduce in other runs.

michaelrsweet:
Hard to diagnose this without a backtrace...

michaelrsweet (changes requested):
Need more information before we can decide this is an appropriate set of fixes.

zdohnal:
[Full backtrace of the crash](https://github.com/zdohnal/pappl-retrofit/blob/fedora-package/gdb-bt) - the program crashes in libcups, because it tries to access invalid memory.

pappl/printer-driver.c
```diff
@@ -968,7 +968,11 @@ make_attrs(
   // printer-device-id
   if (printer->device_id)
   {
-    ippAddString(printer->attrs, IPP_TAG_PRINTER, IPP_TAG_TEXT, "printer-device-id", NULL, printer->device_id);
+    ippAddString(attrs, IPP_TAG_PRINTER, IPP_TAG_TEXT, "printer-device-id", NULL, printer->device_id);
+  }
+  else if ((attr = ippFindAttribute(printer->attrs, "printer-device-id", IPP_TAG_PRINTER)) != NULL)
```
michaelrsweet, line 973: After looking through this, we don't want to look for printer-device-id in `printer->attrs` since it shouldn't be there and you'd end up with duplicate printer-device-id attrs.

pappl/printer-driver.c
```diff
@@ -986,7 +990,10 @@ make_attrs(
     else
       mdl = mfg;			// No separator, so assume the make and model are the same
 
-    formats = ippFindAttribute(printer->driver_attrs, "document-format-supported", IPP_TAG_MIMETYPE);
+    if ((formats = ippFindAttribute(printer->attrs, "document-format-supported", IPP_TAG_MIMETYPE)) == NULL)
```
michaelrsweet, line 993: This should just be looking in attrs since printer->attrs won't have document-format-supported (driver-specific) and we just want document-format-supported since document-format-default is just the one default and usually application/octet-stream.

michaelrsweet (changes requested):
OK, I've had a chance to go through this and yes there are a couple bugs here but the fixes you've proposed aren't quite right. I'll be pushing the correct fixes shortly to both the master and v1.4.x branches. But thank you for catching this!

michaelrsweet:
[master 47cfc65] Fix printer-device-id (Issue #322)

[v1.4.x 8820820] Bump version/copyright.

zdohnal:
Thank you for the fix, Mike! It works in my testing env.

PR #324, opened 2024-01-24
Add PAPPL_API_VERSION_MAJOR/MINOR defines to base.h

proski:
Users of pappl 1.4.5 and newer should be able to access the API version information.

### DIFF
--- a/pappl/base.h
+++ b/pappl/base.h
@@ -33,6 +33,14 @@ extern "C" {
 
 
 //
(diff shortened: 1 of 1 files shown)


michaelrsweet:
Not adding this to 1.4.5 since a) that would be a "feature" change in a bug fix release, and b) what would be the point since 1.4 is the end of the 1.x feature releases.

PR #332, opened 2024-02-06
Fix darkness calculation in web interface

proski:
Round stored percent values to the closest hardware darkness value.


proski:
Verified saving every percentage value in the web interface. All values are preserved on save. Without this change, all values slide towards 0% on save.

michaelrsweet:
Old code, current v1.4.x branch uses master code to calculate.

PR #333, opened 2024-02-06
Fix darkness calculation in web interface

proski:
Round stored percent values to the closest hardware darkness value.

### DIFF
--- a/pappl/printer-webif.c
+++ b/pappl/printer-webif.c
@@ -485,7 +485,7 @@ _papplPrinterWebDefaults(
   if (data.darkness_supported)
   {
     int	d;				// Darkness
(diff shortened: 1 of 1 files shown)


michaelrsweet (commented):
You can't just round the number in one place.

proski:
The original code would only preserve 0%, 20%, 40%, 60%, 80% and 100% for TSPL driver with 16 darkness levels. Other values would slide one step towards 0% on every save.

michaelrsweet:
OK, I've re-tested and agree.

michaelrsweet:
[master 7a395b9] Round darkness value (Issue #333)

[v1.4.x 8f62251] Round darkness value (Issue #333)

PR #336, opened 2024-02-09
Fix gcc warnings

proski:
`body` was used after `realloc()`. Calculate `bodyptr - body` before `realloc()` and use it instead.

GCC suspects that `printer_name` might be used uninitialized. It cannot be. Initialize `printer_name` anyway.

### DIFF
--- a/pappl/client-webif.c
+++ b/pappl/client-webif.c
@@ -194,6 +194,7 @@ papplClientGetForm(
       if (body_size >= body_alloc)
       {
         char *temp;			// Temporary pointer
+        size_t body_ptr_offset = bodyptr - body;
 
         if (body_alloc >= (2 * 1024 * 1024))
           break;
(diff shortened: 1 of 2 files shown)


pappl/client-webif.c
```diff
@@ -194,6 +194,7 @@ papplClientGetForm(
       if (body_size >= body_alloc)
       {
         char *temp;			// Temporary pointer
+        size_t body_ptr_offset = bodyptr - body;
```
michaelrsweet, line 197: I don't add workarounds for compiler warning bugs. Other tools will complain about this construction so there is no point in making this change.

pappl/mainloop-subcommands.c
```diff
@@ -949,7 +949,7 @@ _papplMainloopShowJobs(
     cups_option_t *options)		// I - Options
 {
   const char	*printer_uri,		// Printer URI
-		*printer_name;		// Printer name
+		*printer_name = NULL;	// Printer name
```
michaelrsweet, line 952: Likewise, I'll get a "value stored but not used" warning from other tools.

michaelrsweet (commented):
Sorry, not going to take these changes. See comments.

PR #338, opened 2024-02-09
Fix netcat hang when it send raw data to raw socket

### LINKED ISSUE #331: Raw socket gets data, but does not send data to printer
**Describe the bug**
Printing via raw socket does not work, because file is not sent to the device. This current behavior was seen when using lprint, but probably all PAPPL based printer applications are affected.

**To Reproduce**
1. start ```lprint server``` with ```-o server-options=raw-socket``` 
2. install a printer via ```http://localhost:8000``` - Zebra GK420t in my case (driver ZPL, 4inch, 203dpi)
3. check services on mDNS, f.e. via ```avahi-browse -avrt``` on Linux:

```
=     lo IPv4 zebra                                         _pdl-datastream._tcp local
   hostname = [fedora-home.local]
   address = [127.0.0.1]
   port = [9101]
   txt = ["Scan=F" "PaperMax=<legal-A4" "Fax=F" "product=(Zebra ZPL 4-inch/203dpi/Direct-Thermal)" "priority=100" "qtotal=1" "txtvers=1" "Duplex=F" "Color=F" "UUID=cbf40622-39e6-3048-6ec7-0e9171137724" "pdl=image/pwg-raster,image/urf,application/vnd.zebra-zpl,application/vnd.lprint-test,image/jpeg,image/png" "note=" "adminurl=http://fedora-home.local:8000/zebra/" "ty=Zebra ZPL 4-inch/203dpi/Direct-Thermal"]
```
4. send file via netcat

```
$ netcat 127.0.0.1 9101 < <zpl file>
```

5. Nothing is printed

**Expected behavior**
Printer-ready file will be printed

**System Information:**
 - OS: Fedora Linux 39
 - Browser Firefox
 - lprint 1.2.0
 - pappl 1.4.4

**Additional context**
Attaching debug log from lprint and packet capture between my machine and printer - first it is attempt to send the file via lprint raw socket, then, after approx. 10s, I sent the same file directly to the printer via netcat. I can see in the log that the file is sent to the destination in both cases, but no activity after that in lprint case.

[lprint.pcap.txt](https://github.com/michaelrsweet/pappl/files/14182557/lprint.pcap.txt)
[lprint-raw.log.txt](https://github.com/michaelrsweet/pappl/files/14182576/lprint-raw.log.txt)

zdohnal:
netcat and similar tools hang when they send data to PAPPL's raw socket. It is because these tools wait for input from the destination, and we never end because update `activity` every time `poll()` returns non-negative value, and the tools never send POLLHUP.

The proposed fix is to update `activity` only when we read from the socket, and break out of the loop after 10s of inactivity.

Fixes #331

### DIFF
--- a/pappl/printer-raw.c
+++ b/pappl/printer-raw.c
@@ -152,7 +152,7 @@ _papplPrinterRunRaw(
 
           activity     = time(NULL);
           sockp.fd     = sock;
-          sockp.events = POLLIN | POLLERR;
+          sockp.events = POLLIN | POLLERR | POLLHUP;
 
           for (;;)
           {
@@ -170,12 +170,13 @@ _papplPrinterRunRaw(
 	        continue;
 	    }
 
-            activity = time(NULL);
-
             if (sockp.revents & POLLIN)
             {
               if ((bytes = recv(sock, buffer, sizeof(buffer), 0)) > 0)
+              {
                 write(job->fd, buffer, (size_t)bytes);
+		activity = time(NULL);
+              }
               else
                 break;
             }
@@ -184,7 +185,7 @@ _papplPrinterRunRaw(
               bytes = -1;
               break;
             }
-            else if (sockp.revents & POLLHUP)
+	    else if ((sockp.revents & POLLHUP) || ((time(NULL) - activity) >= 10))
 	    {
 	      break;
 	    }
--- a/pappl/printer-raw.c
+++ b/pappl/printer-raw.c
@@ -112,33 +112,33 @@ _papplPrinterRunRaw(
       // Got a new connection request, accept from the corresponding listener...
       for (i = 0; i < printer->num_raw_listeners; i ++)
       {
-        if (printer->raw_listeners[i].revents & POLLIN)
-        {
-          time_t	activity;	// Network activity watchdog
-          int		sock;		// Client socket
-          http_addr_t	sockaddr;	// Client address
-          socklen_t	sockaddrlen;	// Length of client address
-          struct pollfd	sockp;		// poll() data for client socket
-          pappl_job_t	*job;		// New print job
-          ssize_t	bytes;		// Bytes read from socket
-          char		buffer[8192];	// Copy buffer
-          char		filename[1024];	// Job filename
+	if (printer->raw_listeners[i].revents & POLLIN)
+	{
+	  time_t	activity;	// Network activity watchdog
+	  int		sock;		// Client socket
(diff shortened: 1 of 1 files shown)


michaelrsweet (changes requested):
netcat doesn't behave like the socket backend or any other JetDirect client - when you are done sending data your data you either close the socket or shutdown the outgoing part of the socket, which will signal PAPPL that things are done. netcat is more like a dumb implementation of telnet...

Also, POLLHUP is output only; from the man page:

```
     POLLHUP        The device or socket has been disconnected.  This flag is
                    output only, and ignored if present in the input events
                    bitmask.  Note that POLLHUP and POLLOUT are mutually
                    exclusive and should never be present in the revents
                    bitmask at the same time.
```

As for only updating activity when there is data to be read, it makes sense so that part of your change at least can be adopted.

Finally, it looks like you have a bunch of gratuitous whitespace changes?

michaelrsweet:
[master a483c72] Track activity only on read (Issue #338)

[v1.4.x 06f8942] Track activity only on read (Issue #338)

zdohnal:
> netcat doesn't behave like the socket backend or any other JetDirect client - when you are done sending data your data you either close the socket or shutdown the outgoing part of the socket, which will signal PAPPL that things are done. netcat is more like a dumb implementation of telnet...

I checked netcat and nc (nmap implementation of netcat iiuc) and both were polling the descriptor for events when the binary looked frozen, and the printer application was in pappl iterating in infinite loop with poll().
I'll check the fix in the latest release and let you know if it helps.

> 
> Also, POLLHUP is output only; from the man page:
> 
> ```
>      POLLHUP        The device or socket has been disconnected.  This flag is
>                     output only, and ignored if present in the input events
>                     bitmask.  Note that POLLHUP and POLLOUT are mutually
>                     exclusive and should never be present in the revents
>                     bitmask at the same time.
> ```

I've added it to requested events only for formal reasons - IIUC 'man poll', POLLHUP is always returned in revents if appeared, I meant it like stressing out we check POLLHUP event later (even if it is ignored in events).

> 
> As for only updating activity when there is data to be read, it makes sense so that part of your change at least can be adopted.
> 
> Finally, it looks like you have a bunch of gratuitous whitespace changes?

Bunch of blocks did not follow the code style (unless you've changed the guidelines to differ from CUPS) in the function - 8 spaces weren't substituted by a tab - so I've updated the code to follow the rule. I made it as a separate commit in the PR, so anyone backporting the real fix won't get unnecessary whitespace changes.

PR #339, opened 2024-02-12
printer-raw.c: Make the inactivity timeout 2s

zdohnal:
60s can be too long for successful transfer - it might be taken as serious performance regression in comparison to raw printing to the device itself or via CUPS.

Let's make it 2s for now - if it causes problems later, we can change it ad-hoc.


michaelrsweet:
Like I said, a real socket printing client closes the socket when it is done sending data. "nc" is a poor man's telnet.

2 seconds is way too short, especially if a lot of processing is happening on the client side to prepare each page (think raw printing for EPSON inkjet printers, which the Gutenprint printer application will want to support...)

PR #340, opened 2024-02-13
List raw sockets during `printers` subcommand if available

zdohnal:
Provide a way how to show path to raw socket via CLI without 3rd part tools.

### DIFF
--- a/pappl/mainloop-subcommands.c
+++ b/pappl/mainloop-subcommands.c
@@ -1162,6 +1162,11 @@ _papplMainloopShowPrinters(
   ipp_t		*request,		// IPP request
 		*response;		// IPP response
   ipp_attribute_t *attr;		// Current attribute
+  ipp_tag_t	value_tag;		// Value tag of IPP attribute
+  const char	*printer_name,		// Printer name
+		*printer_uri,		// Printer URI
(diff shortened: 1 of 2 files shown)


pappl/mainloop-subcommands.c
```diff
@@ -1177,8 +1182,46 @@ _papplMainloopShowPrinters(
...
-    puts(ippGetString(attr, 0, NULL));
+  for (attr = ippGetFirstAttribute(response); attr; attr = ippGetNextAttribute(response))
+  {
+    while (attr != NULL && ippGetGroupTag(attr) != IPP_TAG_PRINTER)
+      attr = ippGetNextAttribute(response);
+
+    if (attr == NULL)
+      break;
+
+    value_tag = IPP_TAG_CUPS_INVALID;
```
michaelrsweet, line 1193: I don't think you need to initialize a default `attr_name` or `value_tag` value. And printer-uri-supported is multi-valued so we might as well show all of the values.

michaelrsweet (changes requested):
See comments. I'd also like to see this under a "--verbose" option so the default just lists the printer names.

michaelrsweet:
Please feel free to re-file once you have addressed my review comments...

PR #341, opened 2024-02-13
List raw sockets during `printers` subcommand if available

zdohnal:
Provide a way how to show path to raw socket via CLI without 3rd part tools.


michaelrsweet:
Not adding a feature to 1.4.x...

PR #342, opened 2024-02-15
client-auth.c: Fix authentication denial for valid logins

zdohnal:
According man pages, `getgrouplist()` always return non-zero number, so we have to handle only the case when user is in more groups than we have static array for.

This happens when you run PAPPL based printer application with ```-o auth-service=password-auth -o admin-group=wheel```.

### DIFF
--- a/pappl/client-auth.c
+++ b/pappl/client-auth.c
@@ -122,9 +122,9 @@ _papplClientIsAuthorizedForGroup(
       char	ubuffer[16384];		// User strings
       int	num_groups;		// Number of authenticated groups, if any
 #  ifdef __APPLE__
-      int	groups[32];		// Authenticated groups, if any
+      int	groups[128];		// Authenticated groups, if any
 #  else
-      gid_t	groups[32];		// Authenticated groups, if any
+      gid_t	groups[128];		// Authenticated groups, if any
 #  endif // __APPLE__
 #endif // !_WIN32
 
@@ -157,13 +157,15 @@ _papplClientIsAuthorizedForGroup(
 	    num_groups = (int)(sizeof(groups) / sizeof(groups[0]));
 
 #  ifdef __APPLE__
-	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups) < 0)
 #  else
-	    if (getgrouplist(username, user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, user->pw_gid, groups, &num_groups) < 0)
 #  endif // __APPLE__
 	    {
-	      papplLogClient(client, PAPPL_LOGLEVEL_ERROR, "Unable to lookup groups for user '%s': %s", username, strerror(errno));
-	      num_groups = 0;
+	      papplLogClient(client, PAPPL_LOGLEVEL_WARN, "User '%s' is in more than %d groups.", username, (int)(sizeof(groups) / sizeof(groups[0])));
+#  ifdef __GLIBC__
+	      num_groups = (int)(sizeof(groups) / sizeof(groups[0]));
+#  endif // __GLIBC__
 	    }
 
             // Check group membership...


michaelrsweet:
From the getgrouplist man page on my Mac:

```
RETURN VALUES
     The getgrouplist() function returns 0 on success.  If the size of the
     group list is too small to hold all the user's groups, getgrouplist()
     returns -1 to indicate failure.  In this case, the group array will be
     filled with as many groups as will fit.
```

OpenBSD's implementation seems to match macOS:

```
RETURN VALUES

The getgrouplist() function returns 0 if successful and -1 if the size of the group list is too small to hold all the user's groups. Here, the group array will be filled with as many groups as will fit.
```

But glibc's implementation is different:

```
RETURN VALUE

       If the number of groups of which user is a member is less than or
       equal to *ngroups, then the value *ngroups is returned.

       If the user is a member of more than *ngroups groups, then
       getgrouplist() returns -1.  In this case, the value returned in
       *ngroups can be used to resize the buffer passed to a further
       call to getgrouplist().
```

The fact that you are seeing issues with 32 groups means we need to increase the size of that array, too.

pappl/client-auth.c
```diff
@@ -159,11 +159,11 @@ _papplClientIsAuthorizedForGroup(
 #  ifdef __APPLE__
 	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups))
 #  else
-	    if (getgrouplist(username, user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, user->pw_gid, groups, &num_groups) == -1)
 #  endif // __APPLE__
 	    {
-	      papplLogClient(client, PAPPL_LOGLEVEL_ERROR, "Unable to lookup groups for user '%s': %s", username, strerror(errno));
-	      num_groups = 0;
+	      papplLogClient(client, PAPPL_LOGLEVEL_WARN, "User in too many groups: '%s', cutting to 32", username);
```
michaelrsweet, line 165: We need a better message here, and one that doesn't hardcode the limit. Something like: "User '%s' is in %d groups but only support up to %d groups."

pappl/client-auth.c
```diff
@@ -159,11 +159,11 @@ _papplClientIsAuthorizedForGroup(
 #  ifdef __APPLE__
 	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups))
```
michaelrsweet, line 160: For consistency, add a "< 0" to the condition here.

pappl/client-auth.c
```diff
@@ -159,11 +159,11 @@ _papplClientIsAuthorizedForGroup(
...
 #  else
-	    if (getgrouplist(username, user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, user->pw_gid, groups, &num_groups) == -1)
```
michaelrsweet, line 162: Make this "< 0" instead of comparing to -1.

michaelrsweet (changes requested):
Go ahead and increase the size of the "groups" array to 256 elements.

zdohnal:
> But glibc's implementation is different:
> 
> ```
> RETURN VALUE
> 
>        If the number of groups of which user is a member is less than or
>        equal to *ngroups, then the value *ngroups is returned.
> 
>        If the user is a member of more than *ngroups groups, then
>        getgrouplist() returns -1.  In this case, the value returned in
>        *ngroups can be used to resize the buffer passed to a further
>        call to getgrouplist().
> ```
> 
> The fact that you are seeing issues with 32 groups means we need to increase the size of that array, too.

We can do that too for sure, but the issue was with the condition - the function probably returns 0 only when user would not be in any group (IMHO it is not possible, if the user exists, so it is not mentioned in man page), so it returned non-zero for my user (positive number of groups, since I'm not in more than 32 groups), so the execution got into the if scope and returned failure.

pappl/client-auth.c
zdohnal: Updated to "User '%s' is in %d groups - supported up to %d groups.", using local constant _PAPPL_MAX_GROUPS. The same constant is used for declaration of array.

pappl/client-auth.c
```diff
@@ -21,6 +21,12 @@
 #endif // HAVE_LIBPAM
 
 
+//
+// Local constants...
+//
+
+#define _PAPPL_MAX_GROUPS	32	// Maximum number of groups user can be member of
```
michaelrsweet, line 28: Don't bother with a constant since it is used in one place.

pappl/client-auth.c
```diff
@@ -122,9 +128,9 @@ _papplClientIsAuthorizedForGroup(
       char	ubuffer[16384];		// User strings
       int	num_groups;		// Number of authenticated groups, if any
 #  ifdef __APPLE__
-      int	groups[32];		// Authenticated groups, if any
+      int	groups[_PAPPL_MAX_GROUPS];		// Authenticated groups, if any
```
michaelrsweet, line 131: Make these `[128]`.

pappl/client-auth.c
```diff
@@ -157,13 +163,13 @@ _papplClientIsAuthorizedForGroup(
...
-	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups) < 0)
 #  else
-	    if (getgrouplist(username, user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, user->pw_gid, groups, &num_groups) < 0)
 #  endif // __APPLE__
 	    {
-	      papplLogClient(client, PAPPL_LOGLEVEL_ERROR, "Unable to lookup groups for user '%s': %s", username, strerror(errno));
-	      num_groups = 0;
+	      papplLogClient(client, PAPPL_LOGLEVEL_WARN, "User '%s' is in %d groups - supported up to %d groups.", username, num_groups, _PAPPL_MAX_GROUPS);
```
michaelrsweet, line 171: Since only glibc sets num_groups to a larger value, just sat "User '%s' is in more than %d groups." and use `(int)(sizeof(groups) / sizeof(groups[0]))`.

pappl/client-auth.c
```diff
@@ -157,13 +163,13 @@ _papplClientIsAuthorizedForGroup(
...
+	      num_groups = _PAPPL_MAX_GROUPS;
```
michaelrsweet, line 172: Make this conditional for glibc:

```c
#  ifdef __GLIBC__
        // glibc's getgrouplist implementation sets  num_groups to the total number of groups instead of the number of group stored...
        num_groups = (int)(sizeof(groups) / sizeof(groups[0]);
#  endif // __GLIBC__
```

michaelrsweet (changes requested):
See comments. We obviously need to do something here but need to be careful because getgrouplist is not POSIX.

pappl/client-auth.c
zdohnal: Done.

pappl/client-auth.c
zdohnal: Applied, thanks!

pappl/client-auth.c
zdohnal: Ok, just as I mentioned - the auth did not fail because I'm in too many groups, but due it returned number of groups where I'm in.

pappl/client-auth.c
```diff
@@ -157,13 +157,15 @@
...
 #  else
-	    if (getgrouplist(username, user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, user->pw_gid, groups, &num_groups) < 0)
 #  endif // __APPLE__
 	    {
-	      papplLogClient(client, PAPPL_LOGLEVEL_ERROR, "Unable to lookup groups for user '%s': %s", username, strerror(errno));
-	      num_groups = 0;
+	      papplLogClient(client, PAPPL_LOGLEVEL_WARN, "User '%s' is in more than %d groups.", username, (int)(sizeof(groups) / sizeof(groups[0])));
+#  ifdef __GLIBC__
+	      num_groups = (int)(sizeof(group) / sizeof(groups[0]));
```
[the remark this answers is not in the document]
zdohnal: Fixed.

pappl/client-auth.c
zdohnal: Applied, thanks! (I should copy+paste...)

michaelrsweet (approved):

zdohnal:
@michaelrsweet the message in Changelog is incorrect - it was not issue about user being in more than 32 groups. Auth failed if user was in any group (the function returned number of groups he's in) with glibc.

PR #343, opened 2024-02-20
client-auth.c: Fix authentication denial for valid logins

zdohnal:
According man pages, `getgrouplist()` always return non-zero number, so we have to handle only the case when user is in more groups than we have static array for.

### DIFF
--- a/pappl/client-auth.c
+++ b/pappl/client-auth.c
@@ -122,9 +122,9 @@ _papplClientIsAuthorizedForGroup(
       char	ubuffer[16384];		// User strings
       int	num_groups;		// Number of authenticated groups, if any
 #  ifdef __APPLE__
-      int	groups[32];		// Authenticated groups, if any
+      int	groups[128];		// Authenticated groups, if any
 #  else
-      gid_t	groups[32];		// Authenticated groups, if any
+      gid_t	groups[128];		// Authenticated groups, if any
 #  endif // __APPLE__
 #endif // !_WIN32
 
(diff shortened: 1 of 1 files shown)


michaelrsweet:
getgrouplist returns 0 or -1 on all BSD-derived implementations. Only glibc returns numbers > 0.

pappl/client-auth.c
```diff
@@ -21,6 +21,12 @@
 #endif // HAVE_LIBPAM
 
 
+//
+// Local constants...
+//
+
+#define _PAPPL_MAX_GROUPS	32	// Maximum number of groups user can be member of
```
michaelrsweet, line 28: No need to make this a constant, it is just used to size one array. And based on our discussions for the corresponding PR for the master branch, we wanted to increase this to 128.

pappl/client-auth.c
```diff
@@ -122,9 +128,9 @@ _papplClientIsAuthorizedForGroup(
       char	ubuffer[16384];		// User strings
       int	num_groups;		// Number of authenticated groups, if any
 #  ifdef __APPLE__
-      int	groups[32];		// Authenticated groups, if any
+      int	groups[_PAPPL_MAX_GROUPS];		// Authenticated groups, if any
```
michaelrsweet, line 131: Just make these `[128]`.

pappl/client-auth.c
```diff
@@ -157,13 +163,13 @@ _papplClientIsAuthorizedForGroup(
...
+	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups) < 0)
 #  else
-	    if (getgrouplist(username, user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, user->pw_gid, groups, &num_groups) < 0)
 #  endif // __APPLE__
 	    {
-	      papplLogClient(client, PAPPL_LOGLEVEL_ERROR, "Unable to lookup groups for user '%s': %s", username, strerror(errno));
-	      num_groups = 0;
+	      papplLogClient(client, PAPPL_LOGLEVEL_WARN, "User '%s' is in %d groups - supported up to %d groups.", username, num_groups, _PAPPL_MAX_GROUPS);
+	      num_groups = _PAPPL_MAX_GROUPS;
```
michaelrsweet, line 172: This should be conditional, e.g.:

```c
#  ifdef __GLIBC__
        // glibc's getgrouplist implementation sets  num_groups to the total number of groups instead of the number of group stored...
        num_groups = (int)(sizeof(groups) / sizeof(groups[0]);
#  endif // __GLIBC__
```

michaelrsweet (changes requested):
Note comments. We obviously need to do something but we need to be careful about our usage here since getgrouplist isn't POSIX.

pappl/client-auth.c
zdohnal: Ok, applied.

pappl/client-auth.c
```diff
@@ -157,13 +163,13 @@ _papplClientIsAuthorizedForGroup(
...
-	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, (int)user->pw_gid, groups, &num_groups) < 0)
 #  else
-	    if (getgrouplist(username, user->pw_gid, groups, &num_groups))
+	    if (getgrouplist(username, user->pw_gid, groups, &num_groups) < 0)
 #  endif // __APPLE__
 	    {
-	      papplLogClient(client, PAPPL_LOGLEVEL_ERROR, "Unable to lookup groups for user '%s': %s", username, strerror(errno));
-	      num_groups = 0;
+	      papplLogClient(client, PAPPL_LOGLEVEL_WARN, "User '%s' is in %d groups - supported up to %d groups.", username, num_groups, _PAPPL_MAX_GROUPS);
```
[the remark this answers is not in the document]
zdohnal: Applied.

pappl/client-auth.c
zdohnal: Applied.

michaelrsweet (approved):

PR #345, opened 2024-02-28
dnssd.c: Enable service registration on loopback

zdohnal:
In case users would like to prevent sharing services from printer applications to local network, restrict it to localhost and let CUPS do the sharing.

This can be done by setting `listen-hostname` in PAPPL API - this prevents accessing the public addresses, but the service is still published on those public addresses. This can be prevented if the machine hostname is changed to localhost, but that's not desired on machines IIUC.

The PR does the following:

- introduced new pappl system member `reghost`, which is used for saving `listen-hostname`,
- new public accessors for that member, `papplSystemSetRegHostName()` and `papplSystemGetRegHostName()` - user can set the member to localhost or to the current hostname
- dnssd functions will check this member, and if it is localhost, it will use loopback index
- in case of Avahi it passes NULL as hostname to let Avahi decide what hostname to use (in case of hostname conflicts - and Avahi forbids using localhost if it is not FQDN)

The result is that if reghost is set to localhost, the service is published on `.local` address, but resolved to loopback because CUPS uses DNS-SD names in URIs.

### DIFF
--- a/pappl/dnssd-private.h
+++ b/pappl/dnssd-private.h
@@ -52,6 +52,7 @@ typedef void *_pappl_dns_sd_t;		// DNS-SD master reference
 extern const char	*_papplDNSSDCopyHostName(char *buffer, size_t bufsize) _PAPPL_PRIVATE;
 extern int		_papplDNSSDGetHostChanges(void) _PAPPL_PRIVATE;
 extern _pappl_dns_sd_t	_papplDNSSDInit(pappl_system_t *system) _PAPPL_PRIVATE;
+extern bool		_papplDNSSDIsLoopback(const char *name) _PAPPL_PRIVATE;
 extern void		_papplDNSSDLock(void) _PAPPL_PRIVATE;
 extern const char	*_papplDNSSDStrError(int error) _PAPPL_PRIVATE;
 extern void		_papplDNSSDUnlock(void) _PAPPL_PRIVATE;
--- a/pappl/dnssd.c
+++ b/pappl/dnssd.c
@@ -26,8 +26,13 @@
 
 static char		pappl_dns_sd_hostname[256] = "";
 					// Current DNS-SD hostname
+#ifdef HAVE_MDNSRESPONDER
 static int		pappl_dns_sd_hostname_changes = 0;
 					// Number of host name changes/collisions
+#elif defined(HAVE_AVAHI)
+static int		pappl_dns_sd_hostname_changes = -1;
+					// Number of host name changes/collisions
+#endif // HAVE_MDNSRESPONDER
 static pthread_mutex_t	pappl_dns_sd_hostname_mutex = PTHREAD_MUTEX_INITIALIZER;
 					// Host name mutex
 #ifdef HAVE_MDNSRESPONDER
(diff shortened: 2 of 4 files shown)


michaelrsweet (changes requested):
See my other comments. No to "reghost". Yes to fixing localhost/loopback advertising.

michaelrsweet (changes requested):
See my comments on the other PR.

michaelrsweet:
OK, so I'm really conflicted on this - after reviewing what the changes seek to do, I really don't want to make them.

We already restrict advertisements to localhost when listening only on localhost, and the only way to discover printers from CUPS is via DNS-SD. So I think the current behavior is correct.

zdohnal:
> OK, so I'm really conflicted on this - after reviewing what the changes seek to do, I really don't want to make them.
> 
> We already restrict advertisements to localhost when listening only on localhost,

I'll check the newest release, but IIRC we don't do it (advertising only with localhost address - 127.0.0.1) unless "server-hostname" is set to "localhost" too - which is bad if the printer application is on machine with CUPS Sharing daemon and it needs to shared to be clients via printer profiles or DNS-SD as IPP Everywhere, when user wants to get features of CUPS Sharing server (f.e. AAA) and do not bypass it by accessing printer application directly (like you could implement CUPS Sharing features in pappl as well, but IMO it keeps more sense to leave to CUPS module).
 

> and the only way to discover printers from CUPS is via DNS-SD. So I think the current behavior is correct.

zdohnal:
Yep, the lprint with the latest PAPPL 1.4.7 - the printer is advertised as IPPS service even on non-localhost addresses:

```
= enp1s0 IPv6 test                                          _ipps._tcp           local
   hostname = [zdohnal-fedora.local]
   address = [fe80::5054:ff:fe64:d058]
   port = [8000]
   txt = ["Scan=F" "PaperMax=<legal-A4" "Fax=F" "product=(Zebra ZPL 4-inch/203dpi/Direct-Thermal)" "mopria-certified=1.3" "priority=0" "qtotal=1" "txtvers=1" "Duplex=F" "Color=F" "TLS=1.2" "URF=V1.5,W8,DEVW8,PQ3-4-5,FN3,IS9,MT5,OB9,IFU0,OFU0,RS203" "UUID=7d10e32c-7860-3f84-7806-71ff650da034" "kind=label" "pdl=image/pwg-raster,image/urf,application/vnd.zebra-zpl,application/vnd.lprint-test,image/jpeg,image/png" "note=" "adminurl=http://zdohnal-fedora.local:8000/test/" "ty=Zebra ZPL 4-inch/203dpi/Direct-Thermal" "rp=ipp/print/test"]
= enp1s0 IPv4 test                                          _ipps._tcp           local
   hostname = [zdohnal-fedora.local]
   address = [192.168.124.162]
   port = [8000]
   txt = ["Scan=F" "PaperMax=<legal-A4" "Fax=F" "product=(Zebra ZPL 4-inch/203dpi/Direct-Thermal)" "mopria-certified=1.3" "priority=0" "qtotal=1" "txtvers=1" "Duplex=F" "Color=F" "TLS=1.2" "URF=V1.5,W8,DEVW8,PQ3-4-5,FN3,IS9,MT5,OB9,IFU0,OFU0,RS203" "UUID=7d10e32c-7860-3f84-7806-71ff650da034" "kind=label" "pdl=image/pwg-raster,image/urf,application/vnd.zebra-zpl,application/vnd.lprint-test,image/jpeg,image/png" "note=" "adminurl=http://zdohnal-fedora.local:8000/test/" "ty=Zebra ZPL 4-inch/203dpi/Direct-Thermal" "rp=ipp/print/test"]
=     lo IPv4 test                                          _ipps._tcp           local
   hostname = [zdohnal-fedora.local]
   address = [127.0.0.1]
   port = [8000]
   txt = ["Scan=F" "PaperMax=<legal-A4" "Fax=F" "product=(Zebra ZPL 4-inch/203dpi/Direct-Thermal)" "mopria-certified=1.3" "priority=0" "qtotal=1" "txtvers=1" "Duplex=F" "Color=F" "TLS=1.2" "URF=V1.5,W8,DEVW8,PQ3-4-5,FN3,IS9,MT5,OB9,IFU0,OFU0,RS203" "UUID=7d10e32c-7860-3f84-7806-71ff650da034" "kind=label" "pdl=image/pwg-raster,image/urf,application/vnd.zebra-zpl,application/vnd.lprint-test,image/jpeg,image/png" "note=" "adminurl=http://zdohnal-fedora.local:8000/test/" "ty=Zebra ZPL 4-inch/203dpi/Direct-Thermal" "rp=ipp/print/test"]
```

Configuration:

```
server-options=multi-queue,web-interface,raw-socket,web-network,web-security,web-tls
server-port=8000
log-level=info
listen-hostname=localhost
```

listen-hostname does prevent requests from public network, but does not influence advertising. The purpose of the PR was to fix the advertising for localhost, so there wouldn't be useless services (IMO a service you can't connect to is useless to user and thus not needed to be advertised) on mDNS, since I wouldn't count on Avahi being much scalable with large number of services, so the fewer the better.

michaelrsweet:
OK, not sure why Avahi doesn't include the loopback interface in the list of addresses for "hostname.local" (mDNSResponder certainly does), but I'll open a new issue to track forcing the system hostname to "localhost" when "listen-hostname" is set to "localhost".

michaelrsweet:
Tracking in #369.

PR #346, opened 2024-02-29
dnssd.c: Enable service registration on loopback only

zdohnal:
In case users would like to prevent sharing services from printer
 applications to local network, restrict it to localhost and let CUPS
 do the sharing.

This can be done by setting listen-hostname in PAPPL API - this prevents accessing the public addresses, but the service is still published on those public addresses. This can be prevented if the machine hostname is changed to localhost, but that's not desired on machines IIUC.

The PR does the following:

- introduced new pappl system member reghost, which is used for saving listen-hostname,
- new public accessors for that member, papplSystemSetRegHostName() and papplSystemGetRegHostName() - user can set the member to localhost or to the current hostname
- dnssd functions will check this member, and if it is localhost, it will use loopback index
- in case of Avahi it passes NULL as hostname to let Avahi decide what hostname to use (in case of hostname conflicts - and Avahi forbids using localhost if it is not FQDN)

The result is that if reghost is set to localhost, the service is published on .local address, but resolved to loopback because CUPS uses DNS-SD names in URIs.

PAPPL 2.x version requires CUPS PR https://github.com/OpenPrinting/cups/pull/902 to have it working.

### DIFF
--- a/pappl/dnssd.c
+++ b/pappl/dnssd.c
@@ -51,12 +51,12 @@ _papplPrinterRegisterDNSSDNoLock(
   const char		*papermax;	// PaperMax string value (legacy)
 
 
-  if (!printer->dns_sd_name || !printer->system->is_running || (printer->system->options & PAPPL_SOPTIONS_NO_DNS_SD))
+  if (!printer->dns_sd_name || !printer->system->is_running || (printer->system->options & PAPPL_SOPTIONS_NO_DNS_SD) || !system->hostname)
     return (false);
 
   papplLogPrinter(printer, PAPPL_LOGLEVEL_DEBUG, "Registering DNS-SD name '%s' on '%s'", printer->dns_sd_name, printer->system->hostname);
 
-  if_index = !strcmp(system->hostname, "localhost") ? CUPS_DNSSD_IF_INDEX_LOCAL : CUPS_DNSSD_IF_INDEX_ANY;
+  if_index = _papplDNSSDIsLoopback(system->hostname) ? CUPS_DNSSD_IF_INDEX_LOCAL : CUPS_DNSSD_IF_INDEX_ANY;
 
   // Get attributes and values for the TXT record...
   color_supported           = ippFindAttribute(printer->driver_attrs, "color-supported", IPP_TAG_BOOLEAN);
@@ -303,7 +303,7 @@ _papplSystemRegisterDNSSDNoLock(
 
   papplLog(system, PAPPL_LOGLEVEL_DEBUG, "Registering DNS-SD name '%s' on '%s'", system->dns_sd_name, system->hostname);
 
-  if_index = !strcmp(system->hostname, "localhost") ? CUPS_DNSSD_IF_INDEX_LOCAL : CUPS_DNSSD_IF_INDEX_ANY;
+  if_index = _papplDNSSDIsLoopback(system->hostname) ? CUPS_DNSSD_IF_INDEX_LOCAL : CUPS_DNSSD_IF_INDEX_ANY;
 
   // Rename the service as needed...
   if (system->dns_sd_collision)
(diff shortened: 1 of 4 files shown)


michaelrsweet (changes requested):
I'm not comfortable separating the "registration hostname" from the "listen hostname".

I support changing things to correctly advertise for the loopback interface but not for separating the hostname configurations.

zdohnal:
@michaelrsweet 

> I'm not comfortable separating the "registration hostname" from the "listen hostname".
> 

Hmm... I meant registration hostname to be exactly the listen hostname - and we have to save the 'listen hostname' somewhere to have a way how to decide which net interface to use for advertising, because ```papplSystemAddListeners()``` saves only ports and socket fds, and ```system->hostname``` serves different purpose (setting machine's hostname).

With your feedback, I guess you mean I can save the listen hostname into system struct in ```papplSystemAddListeners()``` instead of having public functions for setting them - if there is no public functions to access them, it prevents possible separation.

> I support changing things to correctly advertise for the loopback interface but not for separating the hostname configurations.

It currently requires changing machine's hostname to change the scope where the service will be advertised - IMO we have to separate server hostname and listen hostname configurations (if you meant them) and have a way where to save listen hostname, or remove the server hostname functionality to change machine's hostname (keep the hostname only internally and do not change /etc/hostname).

WDYT?

zdohnal:
I've removed public accessors and renamed reghost to listen_hostname - it is set only when ```papplSystemAddListeners()``` is called.

I set the PR to [WIP], until  it is clear how I should change it.

zdohnal:
@michaelrsweet ad saving listen-hostname - Or did you mean to save listen-hostname into ```system->hostname``` in ```papplSystemAddListeners()```? That's another way how I imagine to keep the hostnames in sync, but I'm not sure if it is a right call (yeah, if I assign the pointer directly without calling accessor ```papplSystemSetHostName()```, the machine's hostname is not changed, so it could work, but I'm not sure if it is correct thing to do there).

zdohnal:
Ignore the review request for now, I rewrite it to use ```system->hostname```. Once it is ready, I remove the  [WIP].

zdohnal:
Ready for review.

zdohnal:
@michaelrsweet ping

pappl/dnssd.c
```diff
@@ -460,3 +460,25 @@ dns_sd_system_callback(
...
 }
+
+
+//
+// '_papplDNSSDIsLoopback()' - Find out whether the string means
+//                             localhost
+//
+
+bool
+_papplDNSSDIsLoopback(const char *name)
```
michaelrsweet, line 471: I'm still not clear why we need a function for this...

pappl/system-accessors.c
```diff
@@ -141,6 +141,13 @@ papplSystemAddListeners(
       if (ret)
         system->port = port;
     }
+
```
michaelrsweet, line 144: This isn't where I'd save the hostname... You can have multiple listeners.

pappl/dnssd.c
```diff
@@ -51,12 +51,12 @@ _papplPrinterRegisterDNSSDNoLock(
...
 
 
-  if (!printer->dns_sd_name || !printer->system->is_running || (printer->system->options & PAPPL_SOPTIONS_NO_DNS_SD))
+  if (!printer->dns_sd_name || !printer->system->is_running || (printer->system->options & PAPPL_SOPTIONS_NO_DNS_SD) || !system->hostname)
     return (false);
 
   papplLogPrinter(printer, PAPPL_LOGLEVEL_DEBUG, "Registering DNS-SD name '%s' on '%s'", printer->dns_sd_name, printer->system->hostname);
 
-  if_index = !strcmp(system->hostname, "localhost") ? CUPS_DNSSD_IF_INDEX_LOCAL : CUPS_DNSSD_IF_INDEX_ANY;
+  if_index = _papplDNSSDIsLoopback(system->hostname) ? CUPS_DNSSD_IF_INDEX_LOCAL : CUPS_DNSSD_IF_INDEX_ANY;
```
michaelrsweet, line 59: I'm thinking we should examine the listeners or keep a boolean to track whether we are listening on an externally accessible address.

pappl/system.c
```diff
@@ -677,7 +677,7 @@ papplSystemRun(pappl_system_t *system)	// I - System
       bool		force_dns_sd = system->dns_sd_host_changes != dns_sd_host_changes;
 					// Force re-registration?
 
-      if (force_dns_sd)
+      if (!_papplDNSSDIsLoopback(system->hostname) && force_dns_sd)
```
michaelrsweet, line 680: It isn't about being on the loopback interface since we still need to advertise on localhost.

michaelrsweet (changes requested):
Needs work.

michaelrsweet:
OK, like the 1.4.x changes I'm pushing back on this. If listenhost is "localhost", then we only advertise on the loopback interface. This is the correct behavior.

PR #347, opened 2024-03-05
saving loglevel in state file

TheJayas:
In Hp-Lip-Printer App, previously the log level was not being stored in state file and so whenever printer app was starting, the Log Level get its default value as "Error". But now, the recent log level will be stored in state file and whenever printer app restarts, recent log level will be loaded and default log level will be set to that previous log level, which will help in getting different log information of the instance when app is just starting.

### DIFF
--- a/pappl/system-loadsave.c
+++ b/pappl/system-loadsave.c
@@ -83,16 +83,26 @@ papplSystemLoadState(
       papplSystemSetDNSSDName(system, value);
     else if (!strcasecmp(line, "LogLevel"))
     {
-      if(value[0]=='d')
-      {system->log_level=PAPPL_LOGLEVEL_DEBUG;}
-      else if(value[0]=='i')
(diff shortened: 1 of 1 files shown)


TheJayas:
@michaelrsweet , please review this pull request!

michaelrsweet:
@TheJayas You might try waiting for a response more than 3 hours. I'm not a bot, and I have other responsibilities beyond PAPPL.

WRT saving and loading the current log level, that is something I already pull from the server configuration file (new in PAPPL 1.4) but for a variety of reasons I don't save from within the printer application - the user needs to manually change the configuration file to change the default log level on startup. If we load and save the log level like we do for other server state, we'd either need to stop supporting the log-level option for the server sub-command or keep track of whether the option was specified so that it could override the saved state.

Not going to accept this PR right now, and FWIW there are a bunch of coding style changes you'd need to make before I would, but I'll hold onto it to think about what might be done in the future.

PR #352, opened 2024-03-11
Saving Log Level Info-Re

TheJayas:
This pull request is extension of my previous pull request, which got deleted  by mistake when I was trying to add changes in that pull request.
What changes I'm doing with this PR is that I'm saving the log level information in state file and whenever the hp-lip-printer add is rebuilding, the initial log level will be auto-selected by loading the saved information from previous log level(if saved).
This time I have followed the coding style guidelines properly.


michaelrsweet:
This likewise won't work - the saved log level will always override the command-line option.

PR #353, opened 2024-03-18
Handle cases when auth groups are unset

zdohnal:
Currently, pappl based applications require authentication every time for any request if authentication module is set, even if authentication group(s) is set to `None`.

With this PR, admin can set admin and print group to `None`. Every admin action is allowed if admin group is `None`. For printing jobs, Validate-Job, Print-Job, Create-Job are allowed, Send-Document if the requesting client is owner of the job - value of `requesting-user-name` is used as default value, in case authorization field does not contain it.

This allows smooth printing to the printer application via CUPS when printer application uses PAM module to protect against simple attacks.

### DIFF
--- a/pappl/client-accessors.c
+++ b/pappl/client-accessors.c
@@ -334,3 +334,17 @@ papplClientSetUsername(
       client->username[0] = '\0';
   }
 }
+
+
+//
+// '_papplClientGroupIsEmpty()' - Check if group is empty
+//
+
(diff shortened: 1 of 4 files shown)


michaelrsweet (changes requested):
See comments in PR #354.

michaelrsweet:
Latest master should have this fixed.

PR #354, opened 2024-03-18
Handle cases when auth groups are unset

zdohnal:
Currently, pappl based applications require authentication every time for any request if authentication module is set, even if authentication group(s) is set to None.

With this PR, admin can set admin and print group to None. Every admin action is allowed if admin group is None. For printing jobs, Validate-Job, Print-Job, Create-Job are allowed, Send-Document if the requesting client is owner of the job - value of requesting-user-name is used as default value,
 in case authorization field does not contain it.

This allows smooth printing to the printer application via CUPS when printer application uses PAM module to protect against simple attacks.

### DIFF
--- a/pappl/client-accessors.c
+++ b/pappl/client-accessors.c
@@ -338,3 +338,17 @@ papplClientSetUsername(
       client->username[0] = '\0';
   }
 }
+
+
+//
+// '_papplClientGroupIsEmpty()' - Check if group is empty
+//
+
+bool						// O - `true` if empty, `false` if set
+_papplClientGroupIsEmpty(const char *group)	// I - Group name
+{
+  if (!group || !group[0])
+    return (true);
+
+  return (false);
+}
--- a/pappl/client-auth.c
+++ b/pappl/client-auth.c
@@ -216,6 +216,12 @@ _papplClientIsAuthorizedForGroup(
       return (HTTP_STATUS_BAD_REQUEST);
     }
   }
+  else
+  {
+    // If there is no Authorization field, no group and we are on localhost - continue
+    if (_papplClientGroupIsEmpty(group) && httpAddrIsLocalhost(httpGetAddress(client->http)))
+      return (HTTP_STATUS_CONTINUE);
+  }
 
   // If we get there then we don't have any authorization value we can use...
   return (HTTP_STATUS_UNAUTHORIZED);
--- a/pappl/printer-ipp.c
+++ b/pappl/printer-ipp.c
@@ -785,9 +785,21 @@ bool					// O - `true` on success, `false` on failure
 _papplPrinterIsAuthorized(
     pappl_client_t  *client)		// I - Client
 {
+  const char		*username;	// Current username
+  ipp_attribute_t	*attr;	// requesting-user-name attribute
+
   http_status_t code = _papplClientIsAuthorizedForGroup(client, true, client->printer->print_group, client->printer->print_gid);
 
-  if (code == HTTP_STATUS_CONTINUE && client->job && client->job->username && strcmp(client->username, client->job->username))
+  // In case of local connections, use requesting-user-name attribute as username later if client username is missing
+  if (code == HTTP_STATUS_CONTINUE)
+  {
+    if (!client->username[0] && (attr = ippFindAttribute(client->request, "requesting-user-name", IPP_TAG_NAME)) != NULL)
+      username = ippGetString(attr, 0, NULL);
+    else
+      username = client->username;
+  }
+
+  if (code == HTTP_STATUS_CONTINUE && client->job && client->job->username && strcmp(username, client->job->username))
   {
     // Not the owner, try authorizing with admin group...
     code = _papplClientIsAuthorizedForGroup(client, true, client->system->admin_group, client->system->admin_gid);
--- a/pappl/client-private.h
+++ b/pappl/client-private.h
@@ -57,6 +57,7 @@ extern char		*_papplClientCreateTempFile(pappl_client_t *client, const void *dat
 extern void		_papplClientDelete(pappl_client_t *client) _PAPPL_PRIVATE;
 extern void		_papplClientFlushDocumentData(pappl_client_t *client) _PAPPL_PRIVATE;
 extern const char	*_papplClientGetAuthWebScheme(pappl_client_t *client) _PAPPL_PRIVATE;
+extern bool		_papplClientGroupIsEmpty(const char *group) _PAPPL_PRIVATE;
 extern bool		_papplClientHaveDocumentData(pappl_client_t *client) _PAPPL_PRIVATE;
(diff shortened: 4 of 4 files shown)


pappl/client-ipp.c
```diff
@@ -141,6 +141,9 @@ _papplClientProcessIPP(
       client->printer = NULL;
       client->job     = NULL;
 
+      if ((attr = ippFindAttribute(client->request, "requesting-user-name", IPP_TAG_NAME)) != NULL)
```
michaelrsweet, line 144: No, `client->username` is the AUTHENTICATED username, not whatever the client sends in the "requesting-user-name" attribute. There are places where we want to fallback on "requesting-user-name", but this isn't how to do it.

pappl/client-auth.c
```diff
@@ -64,6 +64,10 @@ papplClientIsAuthorized(
   if (!client)
     return (HTTP_STATUS_BAD_REQUEST);
 
+  if (_papplSystemGroupIsEmpty(client->system->admin_group) &&
```
michaelrsweet, line 67: I'm uncomfortable with this change, need to think about the security implications.

pappl/printer-ipp.c
```diff
@@ -38,6 +38,7 @@ static void		ipp_get_jobs(pappl_client_t *client);
 static void		ipp_get_printer_attributes(pappl_client_t *client);
 static void		ipp_hold_new_jobs(pappl_client_t *client);
 static void		ipp_identify_printer(pappl_client_t *client);
+static bool		ipp_is_printing_op(pappl_client_t *client);
```
michaelrsweet, line 41: No, we don't limit authentication to printing operations.

pappl/printer-ipp.c
```diff
@@ -785,6 +786,10 @@ bool					// O - `true` on success, `false` on failure
 _papplPrinterIsAuthorized(
     pappl_client_t  *client)		// I - Client
 {
+  if (_papplSystemGroupIsEmpty(client->printer->print_group) && httpAddrIsLocalhost(httpGetAddress(client->http)) &&
```
michaelrsweet, line 789: This isn't the place for that sort of detemination.

pappl/printer-ipp.c
```diff
@@ -2365,3 +2370,22 @@ valid_job_attributes(
 
   return (valid);
 }
+
+
+static bool					// O - `true` if the operation is related to print job and valid, `false` if not
```
michaelrsweet, line 2375: ALL printers operations except Get-Printer-Attributes MUST be authenticated when the print group is set.

pappl/system-accessors.c
```diff
@@ -2659,3 +2659,17 @@ copy_filter(_pappl_mime_filter_t *f)	// I - Filter definition
 
   return (newf);
 }
+
+
+//
+// '_papplSystemGroupIsEmpty()' - Check if group is empty
+//
+
```
michaelrsweet, line 2667: Empty or NULL group, NOT the "none" group (which is a valid group name and yes people *do* use that name...)

Also, `_papplSystemGroupIsEmpty` implies it operates on a `pappl_system_t` object. Better to have a different name if this is to be generic or provide the equivalent `_papplPrinterXXX` function for the print group.

michaelrsweet (changes requested):
Needs work.

pappl/client-auth.c
zdohnal: AFAIK it has the same implications as the current default in printer applications - any user can do all admin tasks. But I can add check here whether there is an authentication service and move it to ```_papplClientIsAuthorizedForGroup()```.

pappl/client-ipp.c
zdohnal: Yes, I saw this later in the code - I meant this assignment as a fallback for purpose of this code:

```
 806   if (code == HTTP_STATUS_CONTINUE && client->job && client->job->username && strcmp(client->username, client->job->username))
 807   {
 808     // Not the owner, try authorizing with admin group...
 809     code = _papplClientIsAuthorizedForGroup(client, true, client->system->admin_group, client->system->admin_gid);
 810   }
```

where, even if the user is really owner of the job, ```client->username``` is empty, so the request gets to trying authenticate with admin group (which is kind of off-topic for job related operations IMHO).

What I had in mind is to simulate similar settings as default CUPS has - authentication required for admin tasks, printing operations (except for Send-Document) requires no authentication by default.

Does the fallback make sense in ```_papplClientIsAuthorizedForGroup()``` if we don't find authorization field?

pappl/printer-ipp.c
zdohnal: Ok, now I see the function is combining two things - finding out whether the operation is within the expected set and checking the usernames in case of Send-Document, which should not happen. I basically wanted to simulate the default CUPS behavior for printing related operations, so I implemented it in one function, which was to be used only in this use case.

I've thrown it away for now.

pappl/system-accessors.c
zdohnal: > Empty or NULL group, NOT the "none" group (which is a valid group name and yes people _do_ use that name...)

Ok, never knew it exists :D .

> 
> Also, `_papplSystemGroupIsEmpty` implies it operates on a `pappl_system_t` object. Better to have a different name if this is to be generic or provide the equivalent `_papplPrinterXXX` function for the print group.

I've made it ```_papplClientGroupIsEmpty()``` for now, since it is used when authenticating client request (we don't use printer-private.h in client-auth.c).

pappl/client-accessors.c
```diff
@@ -338,3 +338,17 @@ papplClientSetUsername(
...
   }
 }
+
+
+//
+// '_papplClientGroupIsEmpty()' - Check if group is empty
+//
+
+bool						// O - `true` if empty, `false` if set
+_papplClientGroupIsEmpty(const char *group)	// I - Group name
```
michaelrsweet, line 348: Also not an operation on a client object. Why not just look for a NULL or empty string?

pappl/printer-ipp.c
```diff
@@ -785,9 +785,21 @@ bool					// O - `true` on success, `false` on failure
 _papplPrinterIsAuthorized(
     pappl_client_t  *client)		// I - Client
 {
+  const char		*username;	// Current username
+  ipp_attribute_t	*attr;	// requesting-user-name attribute
+
   http_status_t code = _papplClientIsAuthorizedForGroup(client, true, client->printer->print_group, client->printer->print_gid);
 
-  if (code == HTTP_STATUS_CONTINUE && client->job && client->job->username && strcmp(client->username, client->job->username))
+  // In case of local connections, use requesting-user-name attribute as username later if client username is missing
```
michaelrsweet, line 793: No, we want an authenticated username. The alternative (if authentication is enabled) would be to lookup the UID via peer credentials. I don't want to rely on requesting-user-name if authentication is enabled.

michaelrsweet (changes requested):
See comments.

michaelrsweet:
OK, I think with the following change the "None" group should work as expected:

[v1.4.x 1c387a9] Treat blank group name the same as a NULL group name.

PR #355, opened 2024-03-18
Fix setting groups via web ui

zdohnal:
The template is generated incorrectly, so when form data are sent via POST, it does not contain new values for groups to save.

Additionally, set the default value to `none` and make sure the code does not take the value as a bad value.


michaelrsweet (changes requested):
Prefer that the empty string means "no group" since "none" is a valid group name.

michaelrsweet (approved):

PR #365, opened 2024-08-08
system-loadsave.c: Protect state file from world-readable

zdohnal:
The state file might include user-sensitive data, we should prevent accessing it from others than owner.

### DIFF
--- a/pappl/system-webif.c
+++ b/pappl/system-webif.c
@@ -2587,6 +2587,11 @@ tls_install_file(
     papplLogClient(client, PAPPL_LOGLEVEL_ERROR, "Unable to create file '%s': %s", dst, strerror(errno));
     return (false);
   }
+  else
+  {
+    if (fchmod(cupsFileNumber(dstfile), 0644))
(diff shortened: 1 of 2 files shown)


michaelrsweet (changes requested):
This usage isn't safe. You need to use fchmod right after the file is opened, or wrap the open with umask calls. You also need to not error out if the chmod fails since some filesystems/platforms don't support the operation with all values (think Windows in particular...)

Look at what cupsd does in scheduler/file.c with config/state files...

zdohnal:
@michaelrsweet thx, I'll look into it!

zdohnal:
I have found more files which can be protected, so I've added them as well.

pappl/system-webif.c
```diff
@@ -3308,6 +3324,9 @@ tls_make_certsignreq(
   }
   else if ((fp = cupsFileOpen(keyfile, "w")) != NULL)
   {
+    if (fchmod(cupsFileNumber(fp), 0600))
```
michaelrsweet, line 3327: A signing request isn't security-critical...

michaelrsweet:
[master bf52117] Specify restricted permissions for the state file.

The cert/key/csr code is getting overhauled to use the new CUPS X.509 APIs... (#366)

PR #371, opened 2024-11-03
WIP: PAPPL Scan API Bridging v1.4.x (To be rebased after review)

Kappuccino111:
Opening the work completed for the Scan API in v1.4.x directly. Rebasing will be the final step after the base review and testing are completed. The rebasing process should not take long.


michaelrsweet:
OK, I still need to do a more in-depth code review, but I'd prefer that you use [Mini-XML](https://github.com/michaelrsweet/mxml) for any of the XML support instead of libxml2, mainly because libxml2 is about twice the size of PAPPL itself and is nowhere near as fast as Mini-XML.

Kappuccino111:
@michaelrsweet okay, I will consider that. The XML libraries are primarily for parsing the XML side of eSCL requests, so I will have to probably make some library specific changes.

PR #391, opened 2025-07-22
job-process.c: Check the `printfile_cb` for NULL in `filter_raw()`

zdohnal:
legacy-printer-app does not set `printfile_cb`, which occassionally ends up in crash during printing via `legacy-printer-app submit`. It might be an issue in legacy-printer-app as well, but we should protect the library against crashes.

Related: https://github.com/OpenPrinting/pappl-retrofit/issues/30

### DIFF
--- a/pappl/job-process.c
+++ b/pappl/job-process.c
@@ -1162,7 +1162,7 @@ filter_raw(
 {
   papplJobSetImpressions(job, 1);
 
-  if (!(job->printer->driver_data.printfile_cb)(job, doc_number, options, device))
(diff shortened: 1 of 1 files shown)


michaelrsweet (approved):
OK, but there will be more stuff to do here to prevent the "raw" MIME media type from being reported as well (the print file callback is for "raw" printing...)

michaelrsweet:
[master 7fb338a] Validate that print file callback is set for raw files (Issue #391)

[v1.4.x 9a67015] Validate that print file callback is set for raw files (Issue #391)

zdohnal:
@shawnz see https://github.com/OpenPrinting/pappl-retrofit/issues/41 .

PR #422, opened 2026-06-11
Minor test runner fix

johnbeard:
This is a minor fix for the `make test` test runner:

* The testhttpmon test was called twice by `make test` and testqrcode was not called
* Also the `test.log` file is not .gitignored so it clogs up the Git tree as the executed test results land in-tree.


michaelrsweet (approved):
LGTM

Issue #134, opened 2021-01-18
Scanning: Test suite/driver


michaelrsweet:
Add a test driver and end-to-end test functions to the `testpappl` program:

- [ ] Add `pwg-scanner.c` in the testsuite directory, patterned on `pwg-driver.c`; ideally the driver should return test images with different colors/patterns as appropriate, and auto-associate to
- [ ] Update `pwg-driver.c` to include a new "pwg_mfp" driver that auto-associates with the test scanner driver.
- [ ] Update `testpappl.c` to add a "scan" test to validate scanning.
- [ ] Update `testmainloop.c` to call the new scanner APIs prior to `papplMainloop`.

Issue #133, opened 2021-01-18
Scanning: eSCL support


michaelrsweet:
Add eSCL support to PAPPL; no publicly available specification exists, but here are some online resources:

- <https://testcluster.blogspot.com/2014/03/scanning-from-apple-airprint-airscan.html>
- <https://github.com/Cubix651/escl-scanner>
- <https://github.com/Ordissimo/scangearmp2/issues/7>
- <https://github.com/alexpevzner/sane-airscan>

The latter provides a SANE backend that supports scanning via eSCL and can act as a test client.

Support functions should go in a file named `scanner-escl.c`.

tillkamppeter:
The eSCL specification got actually published a year ago:
https://openprinting.github.io/OpenPrinting-News-May-2021/#driverless-scanning
https://mopria.org/spec-download

Issue #131, opened 2021-01-18
Scanning: command-line


michaelrsweet:
Add scan commands for command-line:

- [ ] Add `papplMainloopSetScannerDrivers` API to preset scanner drivers prior to running `papplMainloop`.
- [ ] Add `scan` sub-command with standard options
- [ ] Add `add-scanner` sub-command to add a scanner

PR #425, opened 2026-07-04
Scanning API Investigative PR 

### LINKED ISSUE #130: Scanning: Scanner object, driver Interface, etc.
In order to support multi-function devices and standalone scanners, PAPPL should provide APIs, interfaces, and objects for scanning, with the goal of providing IPP Scan and/or eSCL protocol endpoints (eSCL is used for AirPrint and Mopria, IPP Scan is a published but otherwise unused standard that has the needed semantics/data elements) that can be linked to a printer object and/or used separately.

Specific sub-tasks:

- [ ] Implement `pappl_scanner_t` object based on `pappl_printer_t` (adapted to the needs of scanning - 90% duplication here)
- [ ] Update `pappl_job_t` object to support print or scan jobs (add a pointer to `pappl_scanner_t`)
- [ ] Implement `pappl_sc_driver_data_t` structure based on `pappl_pr_driver_data_t`, with corresponding callback functions using the `pappl_sc_` prefix. *For the purposes of PAPPL scanning, do not support vendor options.*
- [ ] Implement `pappl_sc_options_t` structure based on `pappl_pr_options_t`. *For the purposes of PAPPL scanning, do not support vendor options.*
- [ ] Add `scanner.h` header with public API and `scanner-private.h` header with private API using appropriately-named header guard macros. These can be patterned after the corresponding printer headers.
- [ ] Add `scanner.c` file that provides the scanner object constructor and destructor.
- [ ] Add `scanner-accessors.c` file that provides the scanner object accessor (get/set) functions, patterned on `printer-accessors.c`
- [ ] Add `scanner-driver.c` file that provides the scan driver interfaces and attributes, patterned on `printer-driver.c`.
- [ ] Add `scanner-webif.c` file that provides scan-specific web pages.
- [ ] Add `job-scan.c` file that provides scan job processing functions.
- [ ] Add `papplPrinterSetScanner` API to set the scanner associated with a printer
- [ ] Add `papplSystemSetScannerDrivers` API to set the list of scanner drivers, callback, etc.
- [ ] Add `pappl_sc_autoadd_cb_t`, `pappl_sc_create_cb_t`, `pappl_sc_driver_cb_t`, and `pappl_scanner_cb_t` callbacks to `system.h`.

Need to think about how to auto-associate scanners in printers.

Kappuccino111:
@michaelrsweet @tillkamppeter 

As discussed in #424, this PR is the initial investigative PR towards upstreaming in PAPPL 2.1. 

When merged (probably in multiple steps), will close #130, #131, #133, #134.


pappl/base.h
```diff
@@ -93,6 +93,14 @@ typedef unsigned char pappl_dither_t[16][16];
 typedef struct pappl_pr_driver_data_s pappl_pr_driver_data_t;
 					// Print driver data
 typedef struct _pappl_job_s pappl_job_t;// Job object
+typedef struct pappl_sc_driver_data_s pappl_sc_driver_data_t;
```
michaelrsweet, line 96: We should probably sort this list of types...

pappl/client.c
```diff
@@ -15,6 +15,8 @@
 // Local functions...
 //
 
+static bool	_papplClientMakeHostField(pappl_client_t *client, char *buffer, size_t bufsize);
```
michaelrsweet, line 18: Local functions should be `lowercase_with_underscores`, per "CONTRIBUTING.md".

pappl/client.c
```diff
@@ -381,6 +383,13 @@ _papplClientProcessHTTP(
 	return (papplClientRespond(client, HTTP_STATUS_NOT_FOUND, NULL, NULL, 0, 0));
 
     case HTTP_STATE_GET :
+        // Check for eSCL requests...
```
michaelrsweet, line 386: This should be a resource callback that gets registered when scanners are created.

pappl/client.c
```diff
@@ -434,6 +443,13 @@ _papplClientProcessHTTP(
 	return (papplClientRespond(client, HTTP_STATUS_NOT_FOUND, NULL, NULL, 0, 0));
 
     case HTTP_STATE_POST :
+        // Check for eSCL requests...
```
michaelrsweet, line 446: This should be a resource callback that gets registered when scanners are created.

pappl/client.c
```diff
@@ -473,6 +489,14 @@ _papplClientProcessHTTP(
 	}
 
     default :
+        // Handle DELETE for eSCL job cancellation...
```
michaelrsweet, line 492: This should be a resource callback that is registered when a scanner is created. And we need to break this out as its own case instead of lumping it into the default case.

Also, we need to make sure that the OPTIONS code reports support for the DELETE method.

pappl/client.c
```diff
@@ -631,10 +655,7 @@ papplClientRespondRedirect(
     // Generate an absolute URL...
     char	url[1024];		// Absolute URL
 
-    if (*path == '/')
-      httpAssembleURI(HTTP_URI_CODING_ALL, url, sizeof(url), httpIsEncrypted(client->http) ? "https" : "http", NULL, client->host_field, client->host_port, path);
-    else
-      httpAssembleURIf(HTTP_URI_CODING_ALL, url, sizeof(url), httpIsEncrypted(client->http) ? "https" : "http", NULL, client->host_field, client->host_port, "/%s", path);
+    _papplClientMakeURL(client, httpIsEncrypted(client->http) ? "https" : "http", path, url, sizeof(url));
```
michaelrsweet, line 658: This is not equivalent code.

pappl/client.c
```diff
@@ -651,6 +672,53 @@ papplClientRespondRedirect(
 }
 
 
+//
```
michaelrsweet, line 675: This looks like a special case of the `papplClientRespondRedirect` code, we should just use that function.

pappl/client.c
```diff
@@ -762,3 +830,79 @@ eval_if_modified(
...
+// IPv6 literal addresses must be wrapped in square brackets when used as a
+// host in a URL.  This function adds them if they are missing.  The result
+// is stored in the supplied buffer.
+//
+// Returns true if the host is an IPv6 literal, false otherwise.
+//
+
+static bool
+_papplClientMakeHostField(
```
michaelrsweet, line 846: This should not be necessary if you are using the libcups HTTP APIs.

pappl/client.c
```diff
@@ -762,3 +830,79 @@ eval_if_modified(
...
+  char	host[256];			// Host field, bracketed for IPv6
+  bool	ipv6;				// Is host an IPv6 literal?
+
+
+  ipv6 = _papplClientMakeHostField(client, host, sizeof(host));
+
+  if (ipv6)
+  {
+    // Manually assemble: scheme://[host]:port/path
+    // httpAssembleURI() would percent-encode the brackets.
```
michaelrsweet, line 898: An IPv6 address literal should be detected by `httpAssembleURI(f)` and handled correctly. This snprintf code is not safe since it does not properly escape the host or path.

pappl/client.h
```diff
@@ -54,6 +54,7 @@ extern bool		papplClientIsEncrypted(pappl_client_t *client) _PAPPL_PUBLIC;
 extern bool		papplClientIsValidForm(pappl_client_t *client, size_t num_form, cups_option_t *form) _PAPPL_PUBLIC;
 
 extern bool		papplClientRespond(pappl_client_t *client, http_status_t code, const char *content_coding, const char *type, time_t last_modified, size_t length) _PAPPL_PUBLIC;
+extern bool		papplClientRespondCreated(pappl_client_t *client, const char *location) _PAPPL_PUBLIC;
```
michaelrsweet, line 57: Again, don't think we need this.

pappl/dnssd.c
```diff
@@ -283,6 +284,204 @@ _papplPrinterUnregisterDNSSDNoLock(
 }
 
 
+//
```
michaelrsweet, line 287: This should be part of the printer's service registration, otherwise the scanner and printer service names can get out of sync.

pappl/job-private.h
```diff
@@ -26,6 +26,8 @@
 // Types and structures...
 //
 
+typedef struct pappl_sc_options_s pappl_sc_options_t;	// Forward declaration for eSCL scan options
```
michaelrsweet, line 29: Isn't this already in "base.h"?

pappl/job-private.h
```diff
@@ -47,6 +49,7 @@ struct _pappl_job_s			// Job data
   cups_rwlock_t		rwlock;			// Reader/writer lock
   pappl_system_t	*system;		// Containing system
   pappl_printer_t	*printer;		// Containing printer
+  pappl_scanner_t	*scanner;		// Containing scanner (NULL for print jobs)
```
michaelrsweet, line 52: Since the focus of PAPPL is printers, I think we should not have a separate scanner object here.

michaelrsweet (changes requested):
I've just looked through about 1/3rd of the changes so far but have some concerns about the direction this work as taken...

The primary focus of scanning support in PAPPL is for multi-function printers. I'm honestly not interested in supporting standalone scanners, particularly because SANE already serves that purpose and PAPPL is the Printer Application Framework....

To that end, I'd like to see the code structured for a printer driver callback to specify scanning functions in the `pappl_pr_driver_data_t` structure, which trigger advertising and handling of scanner-specific functions and attributes. Scan jobs can be co-mingled with print jobs for the `pappl_printer_t` object, and then the print and scan code can just pull the next job of the corresponding type when processing.

Kappuccino111:
Okay, working on these @michaelrsweet

tillkamppeter:
> The primary focus of scanning support in PAPPL is for multi-function printers. I'm honestly not interested in supporting standalone scanners, particularly because SANE already serves that purpose and PAPPL is the Printer Application Framework....

@michaelrsweet My original idea in proposing the concept of Scanner Applications was not only support for scanners in multi-function printers using one combined Printer/Scanner Application but also the possibility to distribute scanner drivers as OS-distribution-independent packages and in a way that they work with sandboxed applications and immutable OS distributions, to allow the scan frontend (app which one uses to scan) and scan backend (scanner driver) being in different sandboxes, like we can already do for printing with Printer Applications. Also the communication protocol between frontend and backend in this case should be a general standard one, as eSCL or IPP Scan. In addition, network scanning should work with a general standard protocol and not with the OS-platform-specific SANE one, following your general design philosophy to use general standards (you replaced CUPS' own printer sharing protocol by DNS-SD for example).

SANE connect scan frontends with scan backends by the backends being libraries dynamically linked with the frontends, which does not work through sandbox boundaries and also does not allow OS-distribution-independent packaging of scanner drivers.

To support multi-function printers with a single application and also as many things Printer Applications and Scanner Applications have in common, like HTTP-based protocol, web interface, IPP-over-USB gadget mode, ... I proposed the implementation in PAPPL and not as a separate, stand-alone framework.

For the future I was also thinking about that then a general standard protocol like eSCL or IPP Scan could replace SANE, getting frontends talking this protocol as a client, either directly to driverless scanners or to Scanner Applications as a server, following your philosophy of using general standard protocols. Already existing SANE drivers for legacy scanners then would be continued to be used tucked away in retro-fitting Scanner Applications, like we already do for legacy printers with retro-fitting Printer Applications.

michaelrsweet:
@tillkamppeter 

> My original idea in proposing the concept of Scanner Applications was not only support for scanners in multi-function printers using one combined Printer/Scanner Application but also the possibility to distribute scanner drivers as OS-distribution-independent packages...

PAPPL is for Printer Applications. SANE's existing drivers and eSCL "adapter" can be packaged up as a "legacy scanner application", but the focus of PAPPL (and what *I* am prepared to support and maintain) is PRINTERS.

Kappuccino111:
Also @michaelrsweet, some of the changes you have requested will probably deviate from the original API path. Do we want to update the scanning issues (primarily #130) to reflect scanning being only a part of the printer API, or should I proceed by making those changes directly in this PR?

michaelrsweet:
@Kappuccino111 There is definitely some cleanup that needs to happen for #130, but go ahead and continue your work and I'll deal with that in parallel. The key point is that I've never seen this effort as supporting standalone scanners, but IPP Scan treats the scanner portion of MFDs as a separate service/object with some shared sub-units. Similarly, if there was an IPP Copy service it would use some of the scanner and some of the printer subunits to do its work, but we've never defined that service nor do I see us doing so...

Kappuccino111:
Great then, I will proceed and then message when I am done with the requested changes.

PR #433, opened 2026-08-11
Lock printers_rwlock when deleting a printer (OCU-01-014)

### LINKED ISSUE #432: OCU-01-014: Printer deletion mutates system -> printers under the wrong lock
Filing this from the 7ASecurity audit of OpenPrinting (finding OCU-01-014, WP2, Medium). 
### The problem

Readers of `system->printers` (e.g. `ipp_get_printers()` and the other enumeration paths) take `system->printers_rwlock`, and the creation path in `system-printer.c` also holds `printers_rwlock` around `cupsArrayAdd()`. But `papplPrinterDelete()` in `pappl/printer.c` removes the printer under `_papplRWLockWrite(system)` (i.e. `system->rwlock`) instead:

```c
_papplRWLockWrite(system);
cupsArrayRemove(system->printers, printer);
_papplRWUnlock(system);
```

Since those are two independent locks, a concurrent reader holding `printers_rwlock` isn't protected against this removal, so it can observe the array mutating mid-iteration and end up with a stale printer pointer.

### Fix

Take `printers_rwlock` around the `cupsArrayRemove()`, matching how the creation path in `system-printer.c` locks the array (both `system->rwlock` and `printers_rwlock`, same order). PR shortly.

AayushKumar26:
### What this fixes

Fixes #432 (audit finding OCU-01-014).

`papplPrinterDelete()` removes the printer from `system->printers` under `_papplRWLockWrite(system)` (`system->rwlock`), but readers like `ipp_get_printers()` ; and the creation path in `system-printer.c` which guard that array with `system->printers_rwlock`. Because those are separate locks, a concurrent reader holding `printers_rwlock` isn't protected against the removal and can end up iterating over a stale printer pointer.

### The fix

Take `printers_rwlock` around the `cupsArrayRemove()`, so deletion synchronises with readers the same way creation does. The lock order (`system->rwlock` then `printers_rwlock`) matches `system-printer.c`, so there's no ordering inversion.

### DIFF
--- a/pappl/printer.c
+++ b/pappl/printer.c
@@ -286,7 +286,9 @@ papplPrinterDelete(
 
   // Remove the printer from the system object...
   _papplRWLockWrite(system);
+  cupsRWLockWrite(&system->printers_rwlock);
   cupsArrayRemove(system->printers, printer);
+  cupsRWUnlock(&system->printers_rwlock);
   _papplRWUnlock(system);
 
   _papplPrinterDelete(printer);


michaelrsweet (approved):
Thanks, will also apply to 1.4.x.

michaelrsweet:
[v1.4.x f4dd64b] Fix missing writer lock when deleting a printer (Issue #433)

AayushKumar26:
> Thanks, will also apply to 1.4.x.

Thanks Michael, glad it was useful. Good to know it's going into 1.4.x as well.

